PR #3, opened 2022-08-10
Rework and rename to `rapids-dependency-file-generator`

ajschmidt8:
This PR adds support for `requirements.txt` files and renames the project from `rapids-env-generator` to `rapids-dependency-file-generator`.

It's probably easiest to review the branch as a whole as opposed to reviewing the individual changes in this PR.


README.md
```diff
@@ -1,138 +1,221 @@
...
+  - [`dependencies` key](#dependencies-key)
+- [How Dependency Lists Are Merged](#how-dependency-lists-are-merged)
+- [Additional CLI Notes](#additional-cli-notes)
+
+## Installation
+
+`rapids-dependency-file-generator` is available on PyPI. To install, run:
+
+```sh
+pip install rapids-dependency-file-generator
```
bdice, line 21: It's awkward that the package name and the CLI command are not the same. I'd vote for `rapids-dependency-file-generator` everywhere. I was also okay with the previous name `rapids-env-generator` (I don't think it sounds conda-specific).

README.md
```diff
@@ -1,138 +1,221 @@
...
+```
 
 ## Usage
 
-By default, when `rapids-env-generator` is run, it will look for a `conda/environments/envs.yaml` file relative to the directory from which it's invoked. `rapids-env-generator` will then generate the corresponding environment files based on the contents of the `envs.yaml` file. By default, the generated environment files will be output to the same directory as the `envs.yaml` file (i.e. `conda/environments`).
+When `rapids-dep-file-generator` is invoked, it will read a `dependencies.yaml` file from the current directory and generate children dependency files. `dependencies.yaml` is intended to be committed to the root directory of repositories. It has specific keys (described below) that enable the bifurcation of dependencies for different CUDA versions, architectures, and dependency file types (i.e. `conda` environment files vs `requirements.txt`). The bifurcated dependency lists are merged according to the description in the [_How Dependency Lists Are Merged_](#how-dependency-lists-are-merged) section below.
```
bdice, line 26: Let's use semantic line breaks here - one sentence per line. Much easier to review/edit (I just wanted to add a period after "vs." but had to suggest something for the entire block.)
```suggestion
When `rapids-dep-file-generator` is invoked, it will read a `dependencies.yaml` file from the current directory and generate children dependency files.
`dependencies.yaml` is intended to be committed to the root directory of repositories.
It has specific keys (described below) that enable the bifurcation of dependencies for different CUDA versions, architectures, and dependency file types (i.e. `conda` environment files vs. `requirements.txt`).
The bifurcated dependency lists are merged according to the description in the [_How Dependency Lists Are Merged_](#how-dependency-lists-are-merged) section below.
```

README.md
```diff
@@ -1,138 +1,221 @@
...
+## `dependencies.yaml` Format
 
-Here is an example of an `envs.yaml` file:
+The `dependencies.yaml` file has three relevant top-level keys: `files`, `channels`, and `dependencies`. These keys are described in detail below.
+
+### `files` Key
+
+The top-level `files` key is responsible for determining the following:
+
+- which types of dependency files should be generated (i.e. `conda` environment files and/or `requirements.txt` files)
```
bdice, line 36:
```suggestion
- which types of dependency files should be generated (i.e. conda `environment.yml` files and/or `requirements.txt` files)
```

README.md
```diff
@@ -1,138 +1,221 @@
...
+    includes:
+      - test
 ```
 
-and
+When `generate: none` is used, the `conda_dir`, `requirements_dir` and `matrix` keys can be ommitted. The use case for `generate: none` is described in the [_Additional CLI Notes_](#additional-cli-notes) section below.
+
+### `channels` Key
+
+The top-level `conda` key specifies the channels that should be included in any generated `conda` environment files.
```
bdice, line 91:
```suggestion
The top-level `channels` key specifies the channels that should be included in any generated `conda` environment files.
```

README.md
```diff
@@ -1,138 +1,221 @@
...
+      test: # arbitrarily named dependency list
+        - pytest
+    x86_64-11.5: # common dependencies specific to x86_64-11.5
+      build:
+        - a_random_x86_115_specific_dep
+  conda: # dependencies specific to conda environment files
+    common:
+      build:
+        - cupy
+        - pip: # supports `pip` key for `conda` environment files
```
bdice, line 135: We shouldn't allow the `pip` keyword for `conda_requirements`, right?

README.md
```diff
@@ -1,138 +1,221 @@
...
+For the `11.5` and `x86_64` matrix combination, the following dependency lists would be merged if they exist:
+
+- `conda_requirements.common.build`
+- `conda_requirements.x86_64-11.5.build`
+- `conda.common.build`
+- `conda.x86_64-11.5.build`
+- `conda_requirements.common.test`
+- `conda_requirements.x86_64-11.5.test`
+- `conda.common.test`
+- `conda.x86_64-11.5.test`
```
bdice, line 181: Sort these so that the lines are more clearly grouped:
```suggestion
- `conda_requirements.common.build`
- `conda_requirements.common.test`
- `conda_requirements.x86_64-11.5.build`
- `conda_requirements.x86_64-11.5.test`
- `conda.common.build`
- `conda.common.test`
- `conda.x86_64-11.5.build`
- `conda.x86_64-11.5.test`
```

src/rapids_dep_file_generator/cli.py
```diff
@@ -0,0 +1,67 @@
...
+def validate_args(args):
+    mutually_exclusive_arg_keys = ["file_key", "generate", "cuda_version", "arch"]
+    mutually_exclusive_arg_values = []
+    for i in range(len(mutually_exclusive_arg_keys)):
+        mutually_exclusive_arg_values.append(
+            getattr(args, mutually_exclusive_arg_keys[i])
+        )
+    if any(mutually_exclusive_arg_values) and not all(mutually_exclusive_arg_values):
+        raise Exception(
+            "The following arguments are mutually exclusive and must be used together:"
```
bdice, line 28: I'm not sure if "mutually exclusive" is the right word for this. "Required to be specified together" or similar fits more closely in my view.
```suggestion
            "The following arguments must be used together:"
```

src/rapids_dep_file_generator/cli.py
```diff
@@ -0,0 +1,67 @@
...
+
+def validate_args(args):
+    mutually_exclusive_arg_keys = ["file_key", "generate", "cuda_version", "arch"]
+    mutually_exclusive_arg_values = []
+    for i in range(len(mutually_exclusive_arg_keys)):
+        mutually_exclusive_arg_values.append(
+            getattr(args, mutually_exclusive_arg_keys[i])
+        )
+    if any(mutually_exclusive_arg_values) and not all(mutually_exclusive_arg_values):
+        raise Exception(
```
bdice, line 27: Strongly avoid raising a plain `Exception`, always pick an error class. Here I'd say `ValueError`.
```suggestion
        raise ValueError(
```

src/rapids_dep_file_generator/cli.py
```diff
@@ -0,0 +1,67 @@
...
+        help="The file type to generate",
+        choices=[str(x) for x in [GeneratorTypes.CONDA, GeneratorTypes.REQUIREMENTS]],
+    )
+    inclusive_group.add_argument(
+        "--cuda_version",
+        help="The CUDA version used for generating the output",
+    )
+    inclusive_group.add_argument(
+        "--arch",
+        help="The architecture version used for generating the output",
```
bdice, line 59: An architecture is not really a version.
```suggestion
        help="The architecture used for generating the output",
```

src/rapids_dep_file_generator/constants.py
```diff
@@ -0,0 +1,30 @@
...
+    "rapidsai-nightly",
+    "dask/label/dev",
+    "conda-forge",
+]
+
+default_conda_dir = "conda/environments"
+default_requirements_dir = "python"
+default_dependency_file_path = "dependencies.yaml"
+
+arch_cuda_key_fmt = lambda arch, cuda_ver: f"{arch}-{cuda_ver}"
```
bdice, line 30: Don't name a lambda. Define a function. (Lambdas should be anonymous.)
```suggestion
def arch_cuda_key_fmt(arch, cuda_version):
    return f"{arch}-{cuda_version}"
```

tests/test_rapids_dep_file_generator.py
```diff
@@ -0,0 +1,59 @@
...
+            {"pip": ["pip_dep1", "pip_dep2"]},
+            {"pip": ["pip_dep1", "pip_dep2"]},
+        ]
+    )
+    assert deduped == ["dep1", {"pip": ["pip_dep1", "pip_dep2"]}]
+
+
+@mock.patch("rapids_dep_file_generator.rapids_dep_file_generator.relpath")
+def test_make_dependency_file(mock_relpath):
+    relpath = "../../dependencies.yaml"
```
bdice, line 29: This overrides the imported name `from os.path import relpath`. I was looking for a reason to say "use `import os.path` and call `os.path.join` instead of `from os.path import join`" and this is the kind of thing that justifies that statement. 😄

README.md
ajschmidt8: Sure, I can update that. I didn't like how verbose `rapids-dependency-file-generator` was, but I agree that consistency is probably better.

README.md
ajschmidt8: This is to account for entries like this:

https://github.com/rapidsai/cudf/blob/a67b718b220b7f722ff5f7a375618662419e5695/conda/environments/cudf_dev_cuda11.5.yml#L81-L83

src/rapids_dep_file_generator/cli.py
ajschmidt8: Whoops, agreed. I meant to put `mutually inclusive`, but I'm not sure that's right either. I'll omit it.

src/rapids_dep_file_generator/cli.py
ajschmidt8: Yup, good catch. This was a copy paste error from the `cuda_version` flag

README.md
bdice: It is verbose, but it's just a tab-complete away in most shells, and the verbosity won't hurt if it's called in a script. The arbitrariness of the abbreviation was also questionable - `rapids-dfg` or `rapids-dep-gen` would have been equally okay, but I prefer the full name.

ajschmidt8:
@bdice, thanks for the comments. I am working on the other changes locally as time permits.

PR #4, opened 2022-08-24
Make `matrix` values generic

ajschmidt8:
This PR adjusts the `rapids-dependency-file-generator` tool to be able to accommodate any arbitrary set of matrix values instead of just `cuda` and `arch` values.

### DIFF
--- a/src/rapids_dependency_file_generator/cli.py
+++ b/src/rapids_dependency_file_generator/cli.py
@@ -5,19 +5,18 @@ import yaml
 import argparse
 
 
-def generate_file_obj(config_file, file_key, file_type, cuda_version, arch):
-    if not (config_file and file_key and file_type and cuda_version and arch):
+def generate_file_obj(config_file, file_key, file_type, matrix):
+    if not (config_file and file_key and file_type and matrix):
         return {}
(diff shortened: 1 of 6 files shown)


src/rapids_dependency_file_generator/cli.py
```diff
@@ -38,28 +45,23 @@ def main():
...
-    inclusive_group.add_argument(
-        "--cuda_version",
-        help="The CUDA version used for generating the output",
-    )
-    inclusive_group.add_argument(
-        "--arch",
-        help="The architecture used for generating the output",
+    codependent_args.add_argument(
+        "--matrix",
+        help='string representing which matrix combination should be generated. i.e. --matrix "cuda=11.5;arch=x86_64"',
```
bdice, line 60: This should be `e.g.` and not `i.e.` but it might be better to write `such as`.
```suggestion
        help='string representing which matrix combination should be generated, such as `--matrix "cuda=11.5;arch=x86_64"`',
```

bdice (approved):
Minor suggestions. Thanks @ajschmidt8, I think this is much cleaner now that we have a fully arbitrary matrix.

PR #8, opened 2022-10-18
Fix dependency file paths

ajschmidt8:
This PR includes some changes to ensure that the generated dependency files are written relative to the `dependencies.yaml` file.

### DIFF
--- a/src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
+++ b/src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
@@ -112,8 +112,8 @@ def should_use_specific_entry(matrix_combo, specific_entry_matrix):
     return True
 
 
-def main(config_file_path, files={}):
-    with open(config_file_path, "r") as f:
+def main(config_file_path, files=None):
+    with open(config_file_path) as f:
         parsed_config = yaml.load(f, Loader=yaml.FullLoader)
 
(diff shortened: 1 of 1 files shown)


src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -112,8 +112,8 @@ def should_use_specific_entry(matrix_combo, specific_entry_matrix):
     return True
 
 
-def main(config_file, files):
-    with open(config_file, "r") as f:
+def main(config_file_path, files={}):
```
bdice, line 115: Avoid mutable defaults. Use `None` instead of `{}` and then write
```python
if files is None:
    files = {}
```
(This blog post may be a little strongly worded but it explains the issues that can arise.)
https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -112,8 +112,8 @@ def should_use_specific_entry(matrix_combo, specific_entry_matrix):
...
+    with open(config_file_path, "r") as f:
```
bdice, line 116: Mode `"r"` is the default. The formatting/linting tool `pyupgrade` would remove this automatically.
```suggestion
    with open(config_file_path) as f:
```

PR #10, opened 2022-10-20
Add more tests

ajschmidt8:
This PR includes the following changes:

- Uses `os.makedirs` to ensure that the output directories exist before writing the generated dependency files
- Adds more tests
  - These tests consume a `dependencies.yaml` file to produce some generated output files. The actual generated files are then compared against a set of expected files (which are committed in this PR) to ensure that the results are as expected. The generated files are also added to the `.gitignore` file.


tests/examples/test1/dependencies.yaml
```diff
@@ -0,0 +1,69 @@
+files:
```
bdice, line 1: Maybe we can rename `test1` to have a more descriptive name for what it is testing?

This seems to be a pretty large test file. If we could test the components separately with smaller dependency files, that might be nice. Perhaps this can be `test-integration` and you can add `test-conda-minimal` (a pure test of conda output with a trivial matrix), `test-requirements-minimal` (same for requirements), `test-matrix-minimal` (a minimal conda output with a nontrivial matrix), etc.

tests/examples/test1/dependencies.yaml
```diff
@@ -0,0 +1,69 @@
...
+  all:
+    generate: [conda, requirements]
+    requirements_dir: output/actual
```
bdice, line 4: Perhaps document how to update these tests somewhere: run the generator, manually verify the results, and copy `output/actual` to `output/expected`.

tests/examples/test1/output/expected/all_cuda-115_arch-arm64.yaml
```diff
@@ -0,0 +1,16 @@
+# This file is generated by `rapids-dependency-file-generator`.
```
bdice, line 1: Do we have (or want to add) tests that the generated conda environments are solvable / usable?

tests/test_examples.py
```diff
@@ -0,0 +1,46 @@
...
+
+@pytest.fixture(scope="session", autouse=True)
+def clean_actual_files():
+    for root, _, _ in os.walk("tests"):
+        if os.path.basename(root) == "actual":
+            shutil.rmtree(root)
+
+
+def test1():
+    test_dir = os.path.join(CURRENT_DIR, "examples", "test1")
```
bdice, line 22: You might like to try using `pathlib.Path` objects for most of the places you're using `os.path` in this repo. It's a slightly more modern and concise way of writing things -- but that's an optional refactoring.

tests/test_examples.py
```diff
@@ -0,0 +1,46 @@
...
+    expected_dir = os.path.join(test_dir, "output", "expected")
+    actual_dir = os.path.join(test_dir, "output", "actual")
+    dep_file_path = os.path.join(test_dir, "dependencies.yaml")
+
+    dfg(dep_file_path)
+
+    expected_file_set = {
+        os.path.relpath(f, expected_dir)
```
bdice, line 30: Write a function that takes `expected_dir` and returns the expected files. That would be reusable for `expected` and `actual`.

tests/test_examples.py
bdice, line 22: Use a parametrize here so it's easy to add future tests.
```suggestion
@pytest.mark.parametrize(
    "test_dir",
    ["test1"],
)
def test1(test_dir):
    test_dir = os.path.join(CURRENT_DIR, "examples", test_dir)
```

tests/test_examples.py
```diff
@@ -0,0 +1,46 @@
...
+        if os.path.isfile(f)
+    }
+    actual_file_set = {
+        os.path.relpath(f, actual_dir)
+        for f in glob.glob(actual_dir + "/**", recursive=True)
+        if os.path.isfile(f)
+    }
+
+    assert len(expected_file_set) == len(actual_file_set)
+    assert expected_file_set == actual_file_set
```
bdice, line 41: The length check is kind of redundant -- and if pytest fails at that point, it won't be very helpful because you won't see which files are different. The second check of the sets is helpful and pytest will show nice output if they are not equal. Finally, we want to use the order `assert actual == expected` for the best error messages from pytest. 
```suggestion
    assert actual_file_set == expected_file_set
```

Example: 
image

The order affects how the diff between the two objects is displayed.

tests/examples/test1/output/expected/all_cuda-115_arch-arm64.yaml
ajschmidt8: Eh. That seems kind of difficult and potentially outside the scope of the tool. I think we can skip this for now.

tests/examples/test1/dependencies.yaml
ajschmidt8: Whoops. Forgot to do this in my latest commits. I'll tackle this next.

ajschmidt8: updated in d61da2e

vyasr:
@ajschmidt8 I just got around to reviewing and realized you had merged! In any case everything looks good and I didn't have much to comment on here. The only recommendations I would make (not necessarily in scope for this PR) are:
- We should rename the `main` function in `rapids_dependency_file_generator.py` and reserve that for the CLI entrypoint in `cli.py`.
- We should start adding some documentation soon. Doesn't have to be fancy linted docstrings etc, just some basics for devs to get oriented when they first start looking through.

ajschmidt8:
> @ajschmidt8 I just got around to reviewing and realized you had merged! In any case everything looks good and I didn't have much to comment on here. The only recommendations I would make (not necessarily in scope for this PR) are:
> 
> * We should rename the `main` function in `rapids_dependency_file_generator.py` and reserve that for the CLI entrypoint in `cli.py`.
> * We should start adding some documentation soon. Doesn't have to be fancy linted docstrings etc, just some basics for devs to get oriented when they first start looking through.


Sorry! Just trying to keep things moving since the GH Actions migrations depend on this work. I agree with both of these items.

For the second item, I would argue that we _should_ have these docstrings linted. If RAPIDS is going to depend on this tool, we need to legitimize it as much as possible. I'm not super familiar with Python coding conventions, so if we can programmatically enforce best practices, I think that's a good idea.

PR #13, opened 2022-10-28
Refactor `dependencies.yaml` reading

ajschmidt8:
This PR refactors some things to ensure that the `dependencies.yaml` file is only ever read from the file system a single time.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,53 +1,9 @@
-from unittest import mock
-from unittest.mock import mock_open
-
-from rapids_dependency_file_generator.cli import generate_file_obj, generate_matrix
-
-mock_yaml = """
-files:
-  test:
-    output: none
-    includes:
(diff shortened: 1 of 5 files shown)


tests/test_cli.py
```diff
@@ -1,53 +1,37 @@
...
     matrix = generate_matrix("cuda=11.5;arch=x86_64")
     assert matrix == {"cuda": ["11.5"], "arch": ["x86_64"]}
+
+    matrix = generate_matrix(None)
+    assert matrix == {}
+
+
+def test_validate_args():
+    # Missing output
+    with pytest.raises(Exception):
```
bdice, line 16: Try to use a specific `Exception` subclass like `ValueError`. It's a more specific test and will prevent this from catching `KeyboardInterrupt` like `Ctrl+C` / `^C`.

PR #14, opened 2022-10-31
Update schema

ajschmidt8:
This PR updates the schema for `rapids-dependency-file-generator`.

The top-level `dependencies` key now has key-value pairs beneath it, where the key is equivalent to a name that can be used in the `files.*.includes` lists.


README.md
```diff
@@ -116,60 +116,83 @@ In the absence of a `channels` key, some sensible defaults for RAPIDS will be us
...
 - `specific` - contains dependency lists that are specific to a particular matrix combination
 
-The structure of these two keys varies slightly.
+The values of each of these keys are described in detail below.
+
+#### `common` Key
+
+The `common` key contains a list of objects with the following keys:
+
+- `output_types` - a list of output types (e.g. "conda" for `environment.yaml` files or "requirements" for `requirements.txt` files) for which the packages in the `packages` key should be applied to
```
bdice, line 132: Simplify:
```suggestion
- `output_types` - a list of output types (e.g. "conda" for `environment.yaml` files or "requirements" for `requirements.txt` files) for the packages in the `packages` key
```

README.md
```diff
@@ -116,60 +116,83 @@ In the absence of a `channels` key, some sensible defaults for RAPIDS will be us
...
+- `matrices` - a list of objects (described below) which define packages that are specific to a particular matrix combination
 
-The `common` key has children which are simple key-value pairs, where the key is an arbitrary name for the dependency list (i.e. `build`, `test`, `libcuml_build`, `cuml_build`, etc.) and the value is a list of dependencies.
+##### `matrices` Key
 
-The `specific` key's value is an array of objects. Each object contains a `matrix` key and some arbitrarily named dependency lists (similar to the dependency lists under `common`). The `matrix` key is used to define which matrix combinations from `files.[*].matrix` these dependency lists should apply to. This is elaborated on in [How Dependency Lists Are Merged](#how-dependency-lists-are-merged)
+Each list item under the `matrices` key contains a `matrix` key and a `packages` key. The `matrix` key is used to define which matrix combinations from `files.[*].matrix` these dependency lists should apply to. This is elaborated on in [How Dependency Lists Are Merged](#how-dependency-lists-are-merged). The `packages` key is a list of packages to be included in the generated output file
```
bdice, line 144:
```suggestion
Each list item under the `matrices` key contains a `matrix` key and a `packages` key.
The `matrix` key is used to define which matrix combinations from `files.[*].matrix` will use the associated packages.
The `packages` key is a list of packages to be included in the generated output file for a matching matrix.
This is elaborated on in [How Dependency Lists Are Merged](#how-dependency-lists-are-merged).
```

README.md
```diff
@@ -116,60 +116,83 @@ In the absence of a `channels` key, some sensible defaults for RAPIDS will be us
...
+          - common_build_dep
+      - output_types: conda
+        packages:
+          - cupy
+          - pip: # supports `pip` key for conda environment.yaml files
+              - some_random_dep
+    specific: # dependencies specific to a particular matrix combination
+      - output_types: [conda, requirements]
+        matrices:
+          - matrix: # dependencies specific to x86_64 and 11.5
```
bdice, line 163: Do we need an empty matrix below this, to indicate what happens for `not x86_64 or not cuda==11.5`?

README.md
```diff
@@ -116,60 +116,83 @@ In the absence of a `channels` key, some sensible defaults for RAPIDS will be us
...
+        matrices:
+          - matrix:
+              cuda: "11.5"
+            packages:
+              - another_random_dep=11.5.0
+          - matrix:
+              cuda: "11.6"
+            packages:
+              - another_random_dep=11.6.0
+      - output_types: conda # dependencies specific to conda environment.yaml files
```
bdice, line 178: Maybe put this one higher in the example -- the arch+cuda matrix can be lower.

README.md
```diff
@@ -192,37 +215,55 @@ files:
...
+  - output_types: conda
+    matrices:
+      - matrix:
+          arch: "x86_64"
+        packages:
+          - some_dep1
+          - some_dep2
+```
+
+Every `matrices` list must match exactly one entry (the first match will be used). If no matches are found for a particular matrix combination, an error message will be thrown. In instances where an error should not be thrown, an empty `matrix` and `packages` list item can be used:
```
bdice, line 254:
```suggestion
Every `matrices` list must have a match for a given input matrix (only the first matching matrix in the list of `matrices` will be used).
If no matches are found for a particular matrix combination, an error will be thrown.
In instances where an error should not be thrown, an empty `matrix` and `packages` list item can be used:
```

README.md
```diff
@@ -192,37 +215,55 @@ files:
...
+        cuda: "11.5"
+        arch: x86_64
+        py: "3.8"
+      packages:
+        - a_very_specific_115_x86_38_dep
+    - matrix: # an empty matrix entry serves as a fallback if there are no other matrix matches
+      packages:
 ```
 
 Merged dependency lists are also deduped.
```
bdice, line 269: Are they sorted too? I'd like that if not.
```suggestion
Merged dependency lists are sorted and deduped.
```

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -139,24 +141,38 @@ def make_dependency_files(parsed_config, config_file_path, to_stdout):
...
-                            dependency_entry.get("common", {}).get(include, [])
-                        )
-
-                    for specific_entry in dependency_entry.get("specific", []):
-                        if not should_use_specific_entry(
-                            matrix_combo, specific_entry["matrix"]
-                        ):
+
+                for include in includes:
+                    dependency_entry = parsed_config["dependencies"][include]
```
bdice, line 146: Gives us a free `KeyError` in case of no match! 👍

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -139,24 +141,38 @@ def make_dependency_files(parsed_config, config_file_path, to_stdout):
...
+                        dependencies.extend(common_entry["packages"])
+
+                    for specific_entry in specific_entries:
+                        if file_type not in specific_entry["output_types"]:
                             continue
-                        for include in includes:
-                            dependencies.extend(specific_entry.get(include, []))
+                        specific_matrices = specific_entry["matrices"]
+
+                        foundMatricesMatch = False
```
bdice, line 160: Rename using snake_case.
```suggestion
                        found_matrices_match = False
```

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -139,24 +141,38 @@ def make_dependency_files(parsed_config, config_file_path, to_stdout):
...
+                        for specific_matrices_entry in specific_matrices:
```
bdice, line 161: FYI: You can use a `for...else` in Python. It works like this: https://godbolt.org/z/KqEW19xfG

If you get all the way to the end of the loop without a `break`, then it takes the `else:` clause. With that structure, you don't need a separate variable for `found_matrices_match`.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -139,24 +141,38 @@ def make_dependency_files(parsed_config, config_file_path, to_stdout):
...
+                                continue
+                            foundMatricesMatch = True
+                            dependencies.extend(
+                                specific_matrices_entry["packages"] or []
+                            )
+                            break
+
+                        if not foundMatricesMatch:
+                            raise ValueError(
+                                f"No specific match found in '{include}' for: {matrix_combo}"
```
bdice, line 174:
```suggestion
                                f"No matching matrix found in '{include}' for: {matrix_combo}"
```

tests/examples/specific-fallback/dependencies.yaml
```diff
@@ -0,0 +1,27 @@
...
+        matrices:
+          - matrix:
+              cuda: "11.5"
+            packages:
+              - cudatoolkit=11.5
+          - matrix:
+              cuda: "11.6"
+            packages:
+              - cudatoolkit=11.6
+          - matrix:
```
bdice, line 25: I'd like a test where the empty matrix is at the top, and is always preferred over the other values.

bdice (commented):
Comments attached. Let's address a few of these, then I will approve.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
ajschmidt8: :exploding_head:

README.md
```diff
@@ -116,60 +116,83 @@ In the absence of a `channels` key, some sensible defaults for RAPIDS will be us
...
-  - output_types: [conda] # dependencies specific to conda environment.yaml files
+  build: # dependency list name
+    common: # dependencies common among all matrix variations
+      - output_types: [conda, requirements] # the output types this list item should apply to
+        packages:
+          - common_build_dep
+      - output_types: conda
+        packages:
+          - cupy
+          - pip: # supports `pip` key for conda environment.yaml files
```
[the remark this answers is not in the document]
ajschmidt8: Not sure what you mean here. But now that you raise this question, I realize we don't have any `pip` entries in our tests, so it's probably a good idea to add that.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -139,24 +141,35 @@ def make_dependency_files(parsed_config, config_file_path, to_stdout):
...
+
+                        for specific_matrices_entry in specific_matrices:
+                            if not should_use_specific_entry(
+                                matrix_combo, specific_matrices_entry["matrix"]
+                            ):
+                                continue
+                            dependencies.extend(
+                                specific_matrices_entry["packages"] or []
+                            )
+                            break
```
bdice, line 168: This is a little simpler and doesn't require an explicit `continue`.
```suggestion
                            if should_use_specific_entry(
                                matrix_combo, specific_matrices_entry["matrix"]
                            ):
                                dependencies.extend(
                                    specific_matrices_entry["packages"] or []
                                )
                                break
```

tests/examples/conda-minimal/dependencies.yaml
```diff
@@ -11,29 +11,34 @@ channels:
...
     common:
-      build:
-        - spdlog>=1.8.5,<1.9
+      - output_types: [conda, requirements]
+        packages:
+          - clang=11.1.0
+          - spdlog>=1.8.5,<1.9
+      - output_types: conda
+        packages:
+          - pip:
```
bdice, line 22: I think using a list of `- pip:` requirements typically also requires the actual `pip` package as a dependency.
```suggestion
          - pip
          - pip:
```

bdice (approved):
A couple minor suggestions, otherwise awesome! 🚀

PR #18, opened 2022-11-04
Prevent error from being thrown when no `matrix` values are necessary

### LINKED ISSUE #17: Accept empty `--matrix` values as CLI flag
Currently, if you run the following command:

```sh
  --output conda \
  --file_key test \
  --matrix "" 
```

you'll get an error that says:

```
  File "/home/aj/miniconda3/bin/rapids-dependency-file-generator", line 33, in <module>
    sys.exit(load_entry_point('rapids-dependency-file-generator', 'console_scripts', 'rapids-dependency-file-generator')())
  File "/home/aj/code/nvidia/dependency-file-generator/src/rapids_dependency_file_generator/cli.py", line 63, in main
    args = validate_args(argv)
  File "/home/aj/code/nvidia/dependency-file-generator/src/rapids_dependency_file_generator/cli.py", line 44, in validate_args
    raise ValueError(
ValueError: The following arguments must be used together:
  --file_key
  --output
  --matrix
```

Empty `--matrix` values should probably be allowed. One use case is for environments generated for the `check_style.sh` scripts, where most of the dependencies will just be pure Python dependencies that don't vary per CUDA version, Python version, or architecture.

ajschmidt8:
Fixes #17.

This PR ensures that `rapids-dependency-file-generator` can support generating files that don't require any matrix variations. 

Currently it throws an error if no matrix is provided.


ajschmidt8:
To Do:

- [ ] Update README
- [ ] Update CLI help string?

src/rapids_dependency_file_generator/cli.py
```diff
@@ -39,7 +39,7 @@ def validate_args(argv):
     dependent_arg_keys = ["file_key", "output", "matrix"]
     dependent_arg_values = []
     for key in dependent_arg_keys:
-        dependent_arg_values.append(getattr(args, key))
+        dependent_arg_values.append(getattr(args, key) is None)
```
bdice, line 42: I can't make a suggestion here but we should use a list comprehension instead. Also `getattr(...) is None` is the same as `hasattr(...)`, I believe.
```python
dependent_arg_values = [hasattr(args, key) for key in dependent_arg_keys]
```

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -100,15 +100,18 @@ def get_entry_output_types(output_types):
...
 def get_filename(file_type, file_prefix, matrix_combo):
-    prefix = ""
+    file_type_prefix = ""
     file_ext = ""
     if file_type == str(OutputTypes.CONDA):
         file_ext = ".yaml"
     if file_type == str(OutputTypes.REQUIREMENTS):
         file_ext = ".txt"
-        prefix = "requirements_"
+        file_type_prefix = "requirements_"
```
bdice, line 109: Couple this with the edit below.
```suggestion
        file_type_prefix = "requirements"
```

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -100,15 +100,18 @@ def get_entry_output_types(output_types):
...
     suffix = "_".join([f"{k}-{v}" for k, v in matrix_combo.items()])
-    return f"{prefix}{file_prefix}_{suffix}".replace(".", "") + file_ext
+    filename = "_".join(
+        x for x in [file_type_prefix + file_prefix, suffix] if x
```
bdice, line 112:
```suggestion
        x for x in [file_type_prefix, file_prefix, suffix] if x
```

tests/examples/no-matrix/dependencies.yaml
```diff
@@ -0,0 +1,48 @@
+files:
```
bdice, line 1: Can we minimize the test cases in a future PR? The sample inputs are pretty big and it's hard to pick out what's essential for each test.

bdice (approved):
A few small suggestions -- otherwise LGTM. Good catch, good fix.

src/rapids_dependency_file_generator/cli.py
ajschmidt8: I can change this to a list comprehension. But for the sake of preserving the succinctness of the next line, `if any(dependent_arg_values) and not all(dependent_arg_values):`, I _think_ we need to keep the `is None` logic here to differentiate between an empty string value for `matrix` (e.g. `""`) and it actually being `None`.

If there's a better way to refactor this logic, I'm open to it. I just can't think of any off the top of my head.

tests/examples/no-matrix/dependencies.yaml
ajschmidt8: Yes, we can definitely do that.

src/rapids_dependency_file_generator/cli.py
bdice: Okay. That's fair. Then maybe:

```python
dependent_arg_values = [getattr(args, key) is None for key in dependent_arg_keys]
```

PR #22, opened 2022-11-22
refactor: Migrate data to pyproject.toml

vyasr:
This PR removes setup.py and setup.cfg in favor using a fully PEP 517-compliant build. The linter configs have also moved to pyproject.toml where possible. flake8 does not (and has no plans to) support pyproject.toml, and since it's the only relevant tool doing that I migrated its configuration to a flake8-specific file. I also renamed CONTRIBUTORS.md to CONTRIBUTING.md to better describe what it is (instructions for contributors, not a list of contributors).


vyasr:
If we feel particularly strongly about unifying all config files we could migrate to using [`flakeheaven`](https://github.com/flakeheaven/flakeheaven), which adds pyproject.toml support to `flake8`, but I don't think it's worthwhile. `flake8` is super widely used and I have confidence that the community will find a way to maintain it. I have less confidence in `flakeheaven`.

PR #27, opened 2022-11-23
refactor: Add docstrings and more comments and make minor edits to improve code readability

vyasr:
This PR adds documentation of all functions to explain their purposes and what arguments are being passed around. It also adds more inline comments to explain various different scenarios that might arise, particularly around empty keys or various non-matching cases. Finally, it includes some minor changes to the code that improve readability and remove unnecessary operations.

BREAKING CHANGE: This PR changes the names of a few functions and removes one unused one. This tag is also being used to demarcate our intention to make our first major release when this PR is merged.


ajschmidt8 (approved):
LGTM. Looks great.

Do you think that we should still keep both `specific-fallback-first` and `specific-fallback` tests now that we've made the order of the default value irrelevant? The tests are the same except for the order of the default value. I guess it's useful as a regression test to ensure that we don't break this functionality again in the future. Thoughts?

vyasr:
Yeah no harm in keeping as a regression test IMO.

PR #28, opened 2022-11-28
feat: Support updating conda recipe meta.yaml file

vyasr:
This PR adds support for a new file type to the dependency file generator. The current implementation is really a POC and needs substantial improvements, but it's a good starting point for discussing where the current state of the package and what refactoring is needed to support more general use cases (many of the considerations for meta.yaml will also apply to pyproject.toml).


ajschmidt8:
It seems like this PR generates an entire `meta.yaml` file. In the past, I thought we had talked about just using the existing `conda` output type that already exists for this tool and generating YAML files that can be consumed by `conda build` via the Jinja functions provided below (.e.g. `load_file_data`)

- https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#loading-data-from-other-files

vyasr:
This approach is superseded by #45

PR #29, opened 2022-11-28
Validate dependencies.yaml using jsonschema

vyasr:
This PR enables validating the contents of a dependencies.yaml file directly without doing any processing. The schema is encoded using [JSON Schema](https://json-schema.org/) and validated using [the Python implementation](https://python-jsonschema.readthedocs.io/). The new Python code is fairly minimal, and it would be even shorter except that I leveraged the object-oriented API to show all errors in a file instead of simply showing the first error using `jsonschema.validate`. The vast majority of the new lines are from the schema definition. If we end up merging these changes, then I would recommend moving the schema into a separate JSON file rather than encoding it directly in a Python module, but the current implementation is a sufficient POC.

Whether or not we merge this PR, we should consider updating our schema to change all keys that accept either single strings or arrays of strings to always require the arrays. That would simplify both the schema and our actual parsing code.

### DIFF
--- a/src/rapids_dependency_file_generator/cli.py
+++ b/src/rapids_dependency_file_generator/cli.py
@@ -5,6 +5,7 @@ import yaml
 from ._version import __version__ as version
 from .constants import OutputTypes, default_dependency_file_path
 from .rapids_dependency_file_generator import make_dependency_files
+from .rapids_dependency_file_validator import validate_dependencies
 
 
 def validate_args(argv):
@@ -15,6 +16,11 @@ def validate_args(argv):
         "--config",
         default=default_dependency_file_path,
         help="Path to YAML config file",
+    )
+    parser.add_argument(
+        "--validate",
+        default=True,
+        help="Validate the config file in full before generating anything.",
     )
 
     codependent_args = parser.add_argument_group("optional, but codependent")
@@ -62,6 +68,9 @@ def main(argv=None):
 
     with open(args.config) as f:
         parsed_config = yaml.load(f, Loader=yaml.FullLoader)
+
+    if args.validate:
+        validate_dependencies(parsed_config)
 
     matrix = generate_matrix(args.matrix)
     to_stdout = all([args.file_key, args.output, args.matrix is not None])
--- a/src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
+++ b/src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
@@ -1,7 +1,8 @@
 """Logic for validating dependency files."""
 
+import textwrap
+
 import jsonschema
-import yaml
 
 SCHEMA = {
     "$schema": "http://json-schema.org/draft-07/schema#",
@@ -132,8 +133,22 @@ SCHEMA = {
 }
 
 
-fn = "/home/nfs/vyasr/local/dependency-file-generator/tests/examples/no-specific-match/dependencies.yaml"
-with open(fn) as f:
-    dependency_data = yaml.load(f, Loader=yaml.FullLoader)
+def validate_dependencies(dependencies):
+    """Valid a dictionary against the dependencies.yaml spec.
 
-jsonschema.validate(dependency_data, schema=SCHEMA)
+    Parameters
+    ----------
+    dependencies : dict
+        The parsed dependencies.yaml file.
+
+    Raises
+    ------
+    jsonschema.exceptions.ValidationError
+        If the dependencies do not conform to the schema
+    """
+    validator = jsonschema.Draft7Validator(SCHEMA)
+    if not validator.is_valid(dependencies):
+        for i, error in enumerate(validator.iter_errors(dependencies), start=1):
+            print(f"Error #{i}:")
+            print(textwrap.indent(str(error), "\t"))
+        raise RuntimeError("The provided dependencies data is invalid.")
--- a/src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
+++ b/src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
@@ -1,0 +1,139 @@
+"""Logic for validating dependency files."""
+
+import jsonschema
+import yaml
+
+SCHEMA = {
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://github.com/rapidsai/dependency-file-generator/schema.json",
+    "type": "object",
+    "title": "Dependencies",
+    "description": "A list of all dependencies for a project.",
+    "properties": {
+        "files": {
+            "type": "object",
+            "patternProperties": {
+                ".*": {
+                    "type": "object",
+                    "properties": {
+                        "output": {
+                            "type": ["string", "array"],
+                            "if": {"type": "array"},
+                            "then": {"items": {"type": "string"}},
+                        },
+                        "conda_dir": {"type": "string"},
+                        "requirements_dir": {"type": "string"},
+                        "matrix": {"type": "object"},
+                        "includes": {"type": "array", "items": {"type": "string"}},
+                    },
+                    "required": ["output", "includes"],
+                }
+            },
+        },
+        "channels": {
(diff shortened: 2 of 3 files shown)


vyasr:
As an example, applying this patch:
```
--- a/tests/examples/no-specific-match/dependencies.yaml
+++ b/tests/examples/no-specific-match/dependencies.yaml
@@ -1,14 +1,11 @@
 files:
   all:
-    output: conda
     requirements_dir: output/actual
     matrix:
       cuda: ["11.8"]
     includes:
       - cudatoolkit
-channels:
-  - rapidsai
-  - conda-forge
+channels: 1234
 dependencies:
   cudatoolkit:
     specific:
```
and rerunning tests results in
```
------------------------------------------------------------------------------------------ Captured stderr call -------------------------------------------------------------------------------------------
Error #1:
        'output' is a required property

        Failed validating 'required' in schema['properties']['files']['patternProperties']['.*']:
            {'properties': {'conda_dir': {'type': 'string'},
                            'includes': {'items': {'type': 'string'},
                                         'type': 'array'},
                            'matrix': {'type': 'object'},
                            'output': {'if': {'type': 'array'},
                                       'then': {'items': {'type': 'string'}},
                                       'type': ['string', 'array']},
                            'requirements_dir': {'type': 'string'}},
             'required': ['output', 'includes'],
             'type': 'object'}

        On instance['files']['all']:
            {'includes': ['cudatoolkit'],
             'matrix': {'cuda': ['11.8']},
             'requirements_dir': 'output/actual'}
Error #2:
        1234 is not of type 'array', 'string'

        Failed validating 'type' in schema['properties']['channels']:
            {'if': {'type': 'array'},
             'then': {'items': {'type': 'string'}},
             'type': ['array', 'string']}

        On instance['channels']:
            1234
========================================================================================= short test summary info =========================================================================================
FAILED tests/test_examples.py::test_error_examples[no-specific-match] - RuntimeError: The provided dependencies data is invalid.
====================================================================================== 1 failed, 13 passed in 0.53s =======================================================================================
```

vyasr:
Just making some notes of my planned next steps here to help reviewers understand where I was going with this and identify any blind spots I may have:
- I will move the schema out of the Python file into a JSON file so that it can be maintained separately.
- I would like to implement some form of versioning for the schema. The easiest would be to simply tie its version to the package version.
- I would like to consider removing support for values that are either strings or arrays of strings. I'm not convinced of the value-add given that it leads to undesirable complexity in parsing (all the `_ensure_list` nonsense that we have to make sure to do, which is especially problematic because without that you could silently allow unexpected behavior because the `in` operator in Python will do a substring search when the right operand is a string rather than a collection).

src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
```diff
@@ -0,0 +1,155 @@
...
+        The parsed dependencies.yaml file.
+
+    Raises
+    ------
+    jsonschema.exceptions.ValidationError
+        If the dependencies do not conform to the schema
+    """
+    validator = jsonschema.Draft7Validator(SCHEMA)
+    if not validator.is_valid(dependencies):
+        for i, error in enumerate(validator.iter_errors(dependencies), start=1):
```
csadorf, line 152: It looks like [`iter_errors`](https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/protocols/#jsonschema.protocols.Validator.iter_errors) is deprecated. I would need to investigate whether there is an alternative way to achieve this.

csadorf (changes requested):
In addition to the inline-comments, some general recommendations:

1. We should strive for significantly increasing the composability of the schema file. Using `$ref` links, it is possible to define sub-schemas, such as _RequirementStrings_, _RequirementStrings_ (or _RequirementStringList_), and _Requirements_ which could be either one of the these two.
2. As already mentioned, the schema file should be placed in a separate file.
3. We need to think about how the schema is maintained and also how it is published. One of the easiest approaches is to tie it to the package version and then access it directly via the GitHub raw or download endpoint. In this way the version would be directly encoded as part of the URL. The downside of that approach is that the schema version would potentially change even if there are no changes to it. Finally, it would need to be understood that the schema is part of the API (even if the user-documentation does not directly reference it) and must therefor be backwards-compatible unless a release with breaking changes is made.
4. We should also validate the schema itself as part of the tests for this repository.

Overall, I would recommend to refactor this schema to be significantly more modular before actually checking its correctness. I find it extremely hard to read right now and it would take me significantly more effort compared to refactoring it first and then comparing it to the current non-formalized specification as part of the documentation.

src/rapids_dependency_file_generator/cli.py
```diff
@@ -16,6 +17,11 @@ def validate_args(argv):
         default=default_dependency_file_path,
         help="Path to YAML config file",
     )
+    parser.add_argument(
+        "--validate",
+        default=True,
+        help="Validate the config file in full before generating anything.",
+    )
```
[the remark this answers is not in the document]
vyasr: We're going to wait on this until RAPIDS drops 3.8 support since we need to run dfg inside 3.8 environments until then.

src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
vyasr: We clarified this offline: `iter_errors` itself is not deprecated, only calling it with an extra schema is. The usage here is still valid.

src/rapids_dependency_file_generator/schema.json
```diff
@@ -0,0 +1,161 @@
+ {
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://raw.githubusercontent.com/rapidsai/dependency-file-generator/v1.0.0/src/rapids_dependency_file_generator/schema.json",
```
vyasr, line 3: @ajschmidt8 I'm not sure how best to square including a version number in this id with our use of `semantic-release`. See https://github.com/vyasr/dependency-file-generator/pull/1#discussion_r1061750555 for a longer discussion. Could you comment on what you think is the best path forward here?

vyasr:
Thanks to a makeover from @csadorf I think this PR is ready for review. @ajschmidt8 let me know what you think of it.

One minor note, I snuck in an isort bugfix.

src/rapids_dependency_file_generator/schema.json
ajschmidt8: @vyasr. we can do that with https://github.com/semantic-release/git.

We'll have to figure out how to configure it, but it will commit "release assets" back to the repository (with a `[skip ci]` tag) after a release has occurred.

In this case, `schema.json` would be categorized as a "release asset" since it depends on the release version.

vyasr: OK got it. Is that something that you want to have done before this PR, or are you OK with having this entry be slightly out of date until we implement that afterwards? I don't want to balloon this PR by adding that change here.

ajschmidt8: We can add that in the future.

I can work on that if you don't mind opening up an issue for me.

src/rapids_dependency_file_generator/cli.py
```diff
@@ -16,6 +17,11 @@ def validate_args(argv):
         default=default_dependency_file_path,
         help="Path to YAML config file",
     )
+    parser.add_argument(
+        "--validate",
+        action="store_true",
+        help="Validate the config file in full before generating anything.",
+    )
```
ajschmidt8, line 24: It might make sense to always validate instead of requiring a flag.

vyasr: Done.

ajschmidt8 (approved):

src/rapids_dependency_file_generator/schema.json
vyasr: https://github.com/rapidsai/dependency-file-generator/issues/30

PR #32, opened 2023-02-23
feat: Add option to delete preexisting files created by dfg each run

vyasr:
This adds the ability to request that dfg erase any files it previously created before generating new ones. This logic may be used to replace cleanup logic already present in the tests as well as logic currently embedded [in the GHA tool currently running dfg](https://github.com/rapidsai/gha-tools/blob/37f3946294647eadf5b60d8229ac396a4d913b59/tools/rapids-dependency-file-checker#L9-L15). Moving this logic directly into dfg will allow us to use its pre-commit hook as the primary source of truth for how to run it, both locally and in CI.

Contributes to https://github.com/rapidsai/shared-action-workflows/issues/46


vyasr:
Note: I have intentionally not modified the pre-commit hook to always include the `--clean` argument. Even if we do end up wanting that behavior for every repo, I think that's better specified in their pre-commit configs rather than hardcoding it here.

ajschmidt8 (approved):
Approved with one minor suggestion.

Looks great.

PR #33, opened 2023-02-23
feat: Add support for writing pyproject.toml files

vyasr:
This PR adds support for in-place modification of pyproject.toml files. The modification cannot be done out of place in this case because only a subset of the file is dependencies and there is no way for a pyproject.toml file to point to another location for its dependencies.

On the positive side, the changeset here is quite small and easily manageable. OTOH, there are a number of open design questions that should be addressed before merging this PR. Note that some of these considerations are the same as for conda meta.yaml files in #28. I would consider pyproject.toml to be the easier of the two and require solving a subset of the problems needed for meta.yaml. Essentially all of the problems below exist in similar form for those files, so figuring out how pyproject.toml support should look will get us a large part of the way towards enabling meta.yaml.

1. The interpretation of the `files` key of dependencies.yaml is somewhat suspect when used for pyproject.toml because there is no sensible reason to want to rename the file (this is also true for meta.yaml)
2. For the optional dependencies, we need a way to specify the key of the optional dependencies section. I am currently using the file name for this purpose, but this may be problematic because the file name may not be unique i.e. we might need optional test dependencies in pyproject.toml as well as a test_*.yaml env file (This same problem exists for meta.yaml due to the different host/build/run/test sections). Furthermore, I've currently hacked this in by passing the filename key into the `make_dependency_file` function even though that function doesn't need it for anything else because the output filename has already been generated and is one of the inputs here (this is also why one test is currently failing). If we move forward with this approach, that code will need a bit of refactoring.
3. I am currently splitting up the build requirements, the run dependencies, and the optional dependencies as different OutputTypes, which is functional but somewhat contrived as being distinct from the different sections in (2).

**EDIT**
The above concerns have mostly been addressed. We are now using a single output type for all pyproject sections and allowing a new `extras` section that can be used to provide data specific to pyproject.toml, in this case the table and key under which to store a particular dependency list. This addresses points 2 and 3 above. Point 1 is still somewhat unaddressed, but is mitigated by better documenting the behavior.


vyasr:
This can be tested using https://github.com/rapidsai/rmm/pull/1219

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -108,6 +110,31 @@ def make_dependency_file(
...
+        toml_deps.comment(
+            f"This list was generated by `{cli_name}`. To make changes, edit "
+            f"{relative_path_to_config_file} and run `{cli_name}`."
+        )
+        if file_type == str(OutputTypes.PYPROJECT_BUILD):
+            file_contents["build-system"]["requires"] = toml_deps
+        elif file_type == str(OutputTypes.PYPROJECT_DEPENDENCIES):
+            file_contents["project"]["dependencies"] = toml_deps
+        else:
+            # TODO: May want to make sure the optional-dependencies table exists
```
bdice, line 135: Let’s address this TODO. Check for the key (or some other approach, like catching an exception).

bdice (commented):
All looks fine to me — if we want to go this direction and allow inplace edits of a file. Are you happy with tomlkit’s handling of comments, whitespace, etc. in the file?

I would like to see some regression tests or other forms of testing for this draft before approving.

vyasr:
> All looks fine to me — if we want to go this direction and allow inplace edits of a file. Are you happy with tomlkit’s handling of comments, whitespace, etc. in the file?

Yes, at least for the very limited modifications that we need to make (just a few dependency lists) tomlkit is sufficiently format-preserving for our purposes based on my (admittedly limited) testing. I chose it because it is [the official Python recommendation for a style-preserving TOML library](https://docs.python.org/3/library/tomllib.html).

> I would like to see some regression tests or other forms of testing for this draft before approving.

Agreed, I have no plans to merge this without testing. However, what are your thoughts on the questions I posted in the PR? I would like to address those first before writing new tests in case they result in any meaningful changes to APIs etc.

ajschmidt8 (commented):
This looks awesome.

No other comments aside from what Bradley mentioned.

csadorf:
> 1. The interpretation of the `files` key of dependencies.yaml is somewhat suspect when used for pyproject.toml because there is no sensible reason to want to rename the file (this is also true for meta.yaml)

I see three ways to resolve this:

1. (breaking) We do _not_ use the key to determine the prefix, but instead add a "prefix" keyword to the files entries schema.
2. (non-breaking) Add a new top-level category for "files" that don't exactly follow the current pattern.
3. (non-breaking) Recognize that some type of generated files simply do not use the prefix.

I would advocate for solution 3 as the path with least resistance. All we have to do is update the docs to say that the key has meaning for some outputs, because it is used as a prefix. We could also implement some mix of 1 and 3, where if you provide a "prefix" keyword which defaults to the key were needed.

> 2. For the optional dependencies, we need a way to specify the key of the optional dependencies section. I am currently using the file name for this purpose, but this may be problematic because the file name may not be unique i.e. we might need optional test dependencies in pyproject.toml as well as a test_*.yaml env file (This same problem exists for meta.yaml due to the different host/build/run/test sections). Furthermore, I've currently hacked this in by passing the filename key into the `make_dependency_file` function even though that function doesn't need it for anything else because the output filename has already been generated and is one of the inputs here (this is also why one test is currently failing). If we move forward with this approach, that code will need a bit of refactoring.

Why not simply add another property for this purpose? I've slightly modified the [rmm example](https://github.com/rapidsai/rmm/pull/1219) here:

```yaml
  py_test:
    output: pyproject_optional_dependencies
    extra:
        section: test
    includes:
      - test_python
```

> 3. I am currently splitting up the build requirements, the run dependencies, and the optional dependencies as different OutputTypes, which is functional but somewhat contrived as being distinct from the different sections in (2).

Yeah, it seems a bit clunky, how about this:

```yaml
  py_build:
    output: pyproject
    extra:
        type: build
    includes:
      - build
      - cuda-python
  py_run:
    output: pyproject  # defaults to standard dependencies
    includes:
      - run
      - cuda-python
  py_test:
    output: pyproject
    extra:
      type: optional
      section: test
    includes:
      - test_python
```

Can you add an example file that is then also tested against? I think that would also make it easier to understand and address some of the current issues.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -183,6 +210,14 @@ def get_filename(file_type, file_prefix, matrix_combo):
     if file_type == str(OutputTypes.REQUIREMENTS):
         file_ext = ".txt"
         file_type_prefix = "requirements"
+    if file_type in (
```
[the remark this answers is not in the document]
csadorf: Also in some other spots.

vyasr:
I've applied the proposed adjustments. Will need to do a bit more work to update tests (these tests will require copying in preexisting files for modification rather than writing from scratch) then I think we should be good to go.

README.md
```diff
@@ -98,6 +98,28 @@ files:
...
+      - build
+    extras:
+      table: table_name
+      key: key_name
+```
+
+Currently the supported extras by file type are:
+- pyproject.toml
+  - table: The table in pyproject.toml where the dependencies should be written.
+  - key: The key corresponding to the dependency list in `table`.
```
csadorf, line 121: I understand why you chose to make it flexible like this, but I feel like this is not constrained enough since it allows us to generate invalid `pyproject.toml` files. It breaks my basic assumption for this tool that any valid `dependencies.yaml` file should _always_ generate valid output files.

I had suggested a constrained combination of `type: ['build', 'optional', 'standard'] = 'standard'` and `section: str = None` with noted defaults, because it still allows us to validate that `section` is only used in combination with `optional` and guarantees that we build a valid `pyproject.toml` file.

As another argument against this approach is that it requires us to _always_ specify the `extras` for this output type which is not only annoying, but also error-prone.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -127,8 +131,38 @@ def make_dependency_file(
...
+        toml_deps = tomlkit.array()
+        for dep in dependencies:
+            toml_deps.add_line(dep)
+        toml_deps.add_line(indent="")
+        toml_deps.comment(
+            f"This list was generated by `{cli_name}`. To make changes, edit "
+            f"{relative_path_to_config_file} and run `{cli_name}`."
+        )
+
+        # Recursively descend into subtables like "[x.y.z]", creating tables as needed.
```
csadorf, line 150: While valid TOML, are nested tables like this permitted per the pyproject.toml schema?

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -201,9 +235,13 @@ def get_filename(file_type, file_prefix, matrix_combo):
     file_ext = ""
     if file_type == str(OutputTypes.CONDA):
         file_ext = ".yaml"
-    if file_type == str(OutputTypes.REQUIREMENTS):
+    elif file_type == str(OutputTypes.REQUIREMENTS):
         file_ext = ".txt"
         file_type_prefix = "requirements"
+    elif file_type == str(OutputTypes.PYPROJECT):
+        file_ext = ".toml"
+        # Always hardcoded for this output type.
```
csadorf, line 243: Does this comment still make sense?

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -390,4 +433,7 @@ def make_dependency_files(parsed_config, config_file_path, to_stdout):
                     os.makedirs(output_dir, exist_ok=True)
                     file_path = os.path.join(output_dir, full_file_name)
                     with open(file_path, "w") as f:
-                        f.write(contents)
+                        if file_type == str(OutputTypes.PYPROJECT):
+                            tomlkit.dump(contents, f)
```
csadorf, line 437: Would the `to_stdout` path also be split to use `tomlkit.dump()` ?

tests/test_examples.py
```diff
@@ -47,6 +49,16 @@ def test_examples(example_dir):
     actual_dir = example_dir.joinpath("output", "actual")
     dep_file_path = example_dir.joinpath("dependencies.yaml")
 
+    # Copy pyproject.toml files from expected to actual since they are modified in place
```
csadorf, line 52: Would it make sense to expect that if an output path is provided, that the `pyproject.toml` file is copied there?

tests/test_examples.py
```diff
@@ -47,6 +49,16 @@ def test_examples(example_dir):
...
+    for dirpath, _, filenames in os.walk(expected_dir):
+        for filename in filenames:
+            if filename == "pyproject.toml":
+                full_path = pathlib.PurePath(dirpath) / filename
```
csadorf, line 56: Why are you using `PurePath`? We expect the paths to exist, no?

tests/examples/requirements-minimal/output/expected/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
+    "setuptools",
+] # This list was generated by `rapids-dependency-file-generator`. To make changes, edit ../../dependencies.yaml and run `rapids-dependency-file-generator`.
```
csadorf, line 5: Maybe split this comment into two lines if possible? I would also suggest to add a comment at the very top of the script.

csadorf (changes requested):
Thanks for pushing this forward, I have a few questions and suggestions, and one change request.

README.md
vyasr: Those are good points, I'll address them separately.
- I agree that we should constrain the options to avoid invalid choices. I think that is unquestionably an improvement. We should ensure that dependencies.yaml cannot produce an invalid output file.
- While I see the argument that always specifying the extras is somewhat error-prone, I'm not convinced that the tradeoff of inventing a new set of names is better. I would rather use the exact names that match, i.e. I prefer `table` to `type` and using the actual table name (build-system, project, project.optional-dependencies). Since effectively every project will end up specifying all three (definitely the first two) I don't think it makes sense to privilege one table over the others. We can still ensure that only a small set of expected names works.
- I would be fine making the key an optional field that can only be provided for optional dependencies since in the other cases it is fixed.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
vyasr: Yes, certain specific ones are (e.g. project.optional-dependencies). Which ones we allow can be controlled at the validation stage as we're discussing in the other thread.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
vyasr: Yes. For requirements.txt or conda env files dfg prefixes the file name with the key from the `files` section of dependencies.yaml (e.g. `all_cuda*`), but this doesn't make sense for pyproject.toml files. I will update the comment to describe this more explicitly.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
vyasr: Good question. The `to_stdout` code path doesn't really make sense for pyproject.toml files. The use case for conda env files is to generate new environments on demand, whereas pyproject.toml generation is only useful for a static file. There are a few code paths in here that will fail if `to_stdout == True` and any of the output types if `pyproject`. Rather than spending time trying to enable something that we don't need (at least not right now) I will add some error checking to ensure that this combination doesn't occur (the CLI won't permit it but someone could misuse the Python API in its current form.

tests/test_examples.py
vyasr: I'm not sure that I understand, could you elaborate?

The file tree under "actual" should match the file tree under "expected" exactly. If an output path is provided, then it is assumed that the file tree under "expected" will nest pyproject.toml at that path. Does that answer your question?

README.md
csadorf: I have the impression that we agree on the most important points.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
csadorf: Right, I hadn't realized that this is already two nestings because of the top-level "project".

tests/test_examples.py
csadorf: Sorry, I misunderstood the purpose of lines 52-56. I was proposing to have a mechanism that would not actually modify the pyproject.toml in place, but write to a different location (in this case `actual`). But that probably does not make any sense outside of this particular test context.

tests/test_examples.py
vyasr: Yup will change.

tests/examples/requirements-minimal/output/expected/pyproject.toml
vyasr: I wanted to do the same, unfortunately I don't know how to do this with tomlkit. There is only one comment associated with the line. If I add a manual newline, when reading it back tomlkit no longer associates the second line with the same comment, so each run of dfg just keeps adding another copy of the second line. If you're more familiar with the package and know how this might be accomplished any help would be appreciated!

Re: top of file, I didn't want to do that because there are many parts of this file that are dependencies (and are therefore not auto-generated). I did not want to confuse developers editing those components.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
vyasr: Oh I see what you mean. Yes, I will update both the README and this comment accordingly.

tests/examples/requirements-minimal/output/expected/pyproject.toml
csadorf: Sorry, I'm not familiar enough with tomlkit to be able to know that from the top of my head. As a guess, you probably to add a separate `Comment` node instead of adding the comment to the existing node.

I was thinking more of a disclaimer at the top, like `; Parts of this file were auto-generated, please see the inline notes ...`.

README.md
vyasr: I believe that I've implemented what we want here.

tests/examples/requirements-minimal/output/expected/pyproject.toml
vyasr: Unfortunately I didn't find a good way to address this request yet. I'll play with it later if I can, but for the moment I think it's worth moving forward with the existing changeset and we can add that later if we find a way to do so.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -199,14 +237,21 @@ def get_filename(file_type, file_prefix, matrix_combo):
...
+    elif file_type == str(OutputTypes.REQUIREMENTS):
         file_ext = ".txt"
         file_type_prefix = "requirements"
+    elif file_type == str(OutputTypes.PYPROJECT):
+        file_ext = ".toml"
+        # Unlike for files like requirements.txt or conda environment YAML files, which
+        # may be named with additional prefixes (e.g. all_cuda_*) pyproject.toml files
+        # need to have that exact name and are never prefixed.
+        file_name_prefix = "pyproject"
     suffix = "_".join([f"{k}-{v}" for k, v in matrix_combo.items()])
```
csadorf, line 252: Could the suffix still accidentally apply to pyproject files here?

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -199,14 +237,21 @@ def get_filename(file_type, file_prefix, matrix_combo):
...
     filename = "_".join(
-        x for x in [file_type_prefix, file_prefix, suffix] if x
+        x for x in [file_type_prefix, file_name_prefix, suffix] if x
     ).replace(".", "")
```
csadorf, line 255: This can be simplified to:
```python
filename = "_".join(filter(None, (file_type_prefix, file_name_prefix, suffix))).replace(".", "")
```

src/rapids_dependency_file_generator/schema.json
```diff
@@ -156,6 +158,28 @@
...
+                "properties": { "table": { "const": "project.optional-dependencies" } }
+            },
+            "then": {
+                "required": ["key"]
+            },
+            "else": {
+                "not": {
+                    "required": ["key"]
+                }
+            },
```
csadorf, line 181: Properties are not required by default. Is the "else"-block really necessary?

tests/examples/requirements-minimal/output/expected/pyproject.toml
csadorf: That's totally fine.

csadorf (commented):
Looks much better. Just a few more comments.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
vyasr: Nice use of filter, I never remember the None option.

src/rapids_dependency_file_generator/schema.json
vyasr: From some Googling, I discovered that this not-required idiom is not what it sounds like colloquially. It doesn't mean that "key" is not required, it actually _forbids_ the key from existing. I think that is what we want since otherwise dependencies.yaml could produce invalid pyproject.toml files by providing the wrong key for a particular section (e.g. putting "dependencies" in the "build-requires" table instead of "requires").

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
vyasr: Hmm that is a good question. Zooming out, I think what this question is really asking is whether we should prevent the matrix key from being provided in one of the files entries if the output type is pyproject. We probably should be preventing that at the schema level, would you agree?

src/rapids_dependency_file_generator/schema.json
csadorf: Ok, might be good to add an _invalid_ example to test that.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
csadorf: Yes, ideally we would prevent it at the schema level, but to not overcomplicate things, I would suggest to just check for it in the code and move forward.

src/rapids_dependency_file_generator/schema.json
vyasr: Added a test.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
vyasr: Yup I agree with that. I don't want to overburden the schema.

I've added a check in the code and a corresponding test.

csadorf (approved):
LGTM!

tests/examples/pyproject_matrix/dependencies.yaml
```diff
@@ -0,0 +1,26 @@
+files:
+  py_build:
+    output: pyproject
+    pyproject_dir: output/actual
+    matrix:
+      cuda: ["11.5", "11.6"]
```
ajschmidt8, line 6: When I run `dfg` against this file locally, I get the expected `ValueError: matrix entries are not supported for pyproject.toml files.` error, but I notice that this example isn't under `tests/examples/invalid`.

Is this example supposed to work?

ajschmidt8: Oh, I see now. This example is listed in the `test_error_examples` test case.

These tests might warrant some clean up at some point. But we can save that for another day.

ajschmidt8 (approved):
This is awesome! LGTM.

tests/examples/pyproject_matrix/dependencies.yaml
vyasr: Yeah exactly. I added some comments to test_examples.py explaining the difference between invalid and erroneous tests right now, but it could use some more systematic improvement down the line.

ajschmidt8:
After rerunning the deploy workflow below, this new package version is now in our CI images

- https://github.com/rapidsai/ci-imgs/actions/runs/4349218379/jobs/7621273862

PR #38, opened 2023-03-08
fix: Fix conditional statement

ajschmidt8:
There is currently a conditional statement that is incorrectly attempting to index a string instead of retrieving a value from a dictionary.

Instead of accessing `f["output"]`, it needs to be adjusted to `files[f]["output"]`

The code excerpt below shows a minimal reproducer of the problem and the solution.

```python
files = {'checks': {'matrix': {'cuda': ['11.8'], 'arch': ['x86_64'], 'py': ['3.10']}, 'output': 'conda', 'includes': ['checks', 'py_version']}}

# this throws error: "TypeError: string indices must be integers"
for f in files:
  f["output"]

# this fixes it
for f in files:
   files[f]["output"]
```

### DIFF
--- a/src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
+++ b/src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
@@ -351,7 +351,8 @@ def make_dependency_files(parsed_config, config_file_path, to_stdout):
     channels = parsed_config.get("channels", default_channels) or default_channels
     files = parsed_config["files"]
     if to_stdout and any(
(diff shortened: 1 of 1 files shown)


bdice (approved):
Can we add a test for this? Fine if it happens in a follow-up PR, to unblock CI.

ajschmidt8:
> Can we add a test for this? Fine if it happens in a follow-up PR, to unblock CI.

Yup, opened https://github.com/rapidsai/dependency-file-generator/issues/39.

I'll merge this for now.

PR #47, opened 2023-05-19
fix: print schema errors to `stderr`

trxcllnt:


trxcllnt:
Looks like there's an error with the semantic-pr-title workflow from the GHE migration. `amannn/action-semantic-pull-request@*` needs to be added to the allowlist.

PR #54, opened 2023-10-04
fix: Remove dependency on pkg_resources

### LINKED ISSUE #53: pkg_resources is deprecated and removed in Python 3.12
CI failed with the error:

```
ModuleNotFoundError: No module named 'pkg_resources'
```

https://github.com/rapidsai/cuxfilter/actions/runs/6400270892/job/17373650755?pr=542#step:6:172

This is because the style check job used Python 3.12 which removed `setuptools`, `pkg_resources`, and some other libraries by default.

We need to migrate this code to use `importlib.resources` instead:

https://github.com/rapidsai/dependency-file-generator/blob/56ea881c2c3c3f7489073b47ecec6aefbfbdb6e4/src/rapids_dependency_file_generator/rapids_dependency_file_validator.py#L8

bdice:
Closes #53 by replacing use of `pkg_resources` with `importlib.resources`. This should fix Python 3.12 support.

### DIFF
--- a/src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
+++ b/src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
@@ -1,14 +1,16 @@
 """Logic for validating dependency files."""
 
+import importlib
 import json
 import sys
 import textwrap
 
 import jsonschema
-import pkg_resources
 from jsonschema.exceptions import best_match
 
-SCHEMA = json.loads(pkg_resources.resource_string(__name__, "schema.json"))
+SCHEMA = json.loads(
+    importlib.resources.files(__name__).joinpath("schema.json").read_bytes()
+)
 
 
 def validate_dependencies(dependencies):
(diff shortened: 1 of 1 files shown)


src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
```diff
@@ -1,14 +1,16 @@
...
 import textwrap
 
 import jsonschema
-import pkg_resources
 from jsonschema.exceptions import best_match
 
-SCHEMA = json.loads(pkg_resources.resource_string(__name__, "schema.json"))
+SCHEMA = json.loads(
+    importlib.resources.files(__package__).joinpath("schema.json").read_bytes()
+)
```
vyasr, line 13: In case anyone else comes to look at this PR, this change aligns with the official recommendation: https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-string

vyasr (approved):

src/rapids_dependency_file_generator/rapids_dependency_file_validator.py
bdice: Oh nice! I googled for a migration guide but I just guessed what code to use and got it right. 🏆

PR #59, opened 2023-12-13
fix: Avoid crash when file in repo contains non-unicode characters

### LINKED ISSUE #58: Tool fails when using `--clean` if files with non-unicode characters exist in the repo
If you run `rapids-dependency-file-generator --clean` with any file that has non-unicode characters in it, it will generate the following exception:

```
  File "${VIRTUAL_ENV}/bin/rapids-dependency-file-generator", line 8, in <module>
    sys.exit(main())
  File "${VIRTUAL_ENV}/lib/python3.10/site-packages/rapids_dependency_file_generator/cli.py", line 103, in main
    delete_existing_files(args.clean)
  File "${VIRTUAL_ENV}/lib/python3.10/site-packages/rapids_dependency_file_generator/rapids_dependency_file_generator.py", line 40, in delete_existing_files
    if HEADER in f.read():
  File "${VIRTUAL_ENV}/lib/python3.10/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa0 in position 6038: invalid start byte
```

There should be a try block around reading the file to ignore files that can not be decoded. Something like this fixed it for me:

```python
def delete_existing_files(root="."):
    """Delete any files generated by this generator.

    This function can be used to clean up a directory tree before generating a new set
    of files from scratch.

    Parameters
    ----------
    root : str
        The path to the root of the directory tree to search for files to delete.
    """
    for dirpath, _, filenames in os.walk(root):
        for fn in filter(
            lambda fn: fn.endswith(".txt") or fn.endswith(".yaml"), filenames
        ):
            try:
                with open(file_path := os.path.join(dirpath, fn)) as f:
                    if HEADER in f.read():
                        os.remove(file_path)
            except UnicodeDecodeError:
                pass
```

mdemoret-nv:
This fix prevents a crash when using `--clean` on a repo where a file contains non-unicode characters. Fix seems benign since any file with non-unicode characters likely wasnt created by this tool.

Closes #58

### DIFF
--- a/src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
+++ b/src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
@@ -35,12 +35,12 @@ def delete_existing_files(root="."):
         for fn in filter(
             lambda fn: fn.endswith(".txt") or fn.endswith(".yaml"), filenames
         ):
-            try:
-                with open(file_path := os.path.join(dirpath, fn)) as f:
+            with open(file_path := os.path.join(dirpath, fn)) as f:
+                try:
                     if HEADER in f.read():
                         os.remove(file_path)
-            except UnicodeDecodeError:
-                pass
+                except UnicodeDecodeError:
+                    pass
 
 
 def dedupe(dependencies):
--- a/src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
+++ b/src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
@@ -35,9 +35,12 @@ def delete_existing_files(root="."):
         for fn in filter(
             lambda fn: fn.endswith(".txt") or fn.endswith(".yaml"), filenames
         ):
-            with open(file_path := os.path.join(dirpath, fn)) as f:
-                if HEADER in f.read():
-                    os.remove(file_path)
+            try:
+                with open(file_path := os.path.join(dirpath, fn)) as f:
+                    if HEADER in f.read():
+                        os.remove(file_path)
+            except UnicodeDecodeError:
+                pass
 
 
 def dedupe(dependencies):


vyasr:
I agree that this seems safe, but would like confirmation from one more dev just in case there's a use case I'm missing.

bdice:
@mdemoret-nv Can you share an example file where this failed, and also the full error output you got with the `UnicodeDecodeError`?

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -35,9 +35,12 @@ def delete_existing_files(root="."):
...
         ):
-            with open(file_path := os.path.join(dirpath, fn)) as f:
-                if HEADER in f.read():
-                    os.remove(file_path)
+            try:
+                with open(file_path := os.path.join(dirpath, fn)) as f:
+                    if HEADER in f.read():
+                        os.remove(file_path)
+            except UnicodeDecodeError:
+                pass
```
bdice, line 43: I don't think we want to skip cleaning the file in this case. I'm not sure what we want, though. If you can provide an example file, I would like to dig into this and try it locally before approving this approach.

bdice, line 43: Let's swap the scope of the `try` and `with` here, since it's the `f.read()` that is problematic. I spoke with @vyasr offline and we agreed this is a good solution.
```suggestion
            with open(file_path := os.path.join(dirpath, fn)) as f:
                try:
                    if HEADER in f.read():
                        os.remove(file_path)
                except UnicodeDecodeError:
                    pass
```

bdice (approved):
Discussed with @vyasr and this seems like a reasonable fix. I found a way to reproduce this locally, so we're good to merge once my suggestion is applied.

mdemoret-nv:
> Discussed with @vyasr and this seems like a reasonable fix. I found a way to reproduce this locally, so we're good to merge once my suggestion is applied.

Committed. Thanks!

PR #61, opened 2023-12-14
feat: Add support for matrix globs.

### LINKED ISSUE #41: Support matrix globs
Feature proposal: `matrix:` entries should be interpreted as globs, so that `matrix: cuda: "11.*"` can match any of `11.2, 11.4, 11.5, 11.8`. This will be really helpful when delineating CUDA 11 from CUDA 12, so that we don't have to specify every version individually if the match is major-version-only and not specific to minor versions.

We would keep the current behavior of taking the first match. If a specific match is to take precedence, it should be listed before the glob match. e.g. `11.8` before `11.*`.

Example:
```yaml
  cuda-python:
    specific:
      - output_types: conda
        matrices:
          - matrix:
              cuda: "11.*"
            packages:
              - cuda-python>=11.7.0
          - matrix:
              cuda: "12.*"
            packages:
              - cuda-python>=12.1.0
```

bdice:
Closes #41.


src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -321,7 +323,8 @@ def should_use_specific_entry(matrix_combo, specific_entry_matrix):
         `matrix_combo` and False otherwise.
     """
     return all(
-        matrix_combo.get(specific_key) == specific_value
+        specific_key in matrix_combo
+        and fnmatch.fnmatch(matrix_combo.get(specific_key), specific_value)
```
vyasr, line 327:
```suggestion
        and fnmatch.fnmatch(matrix_combo[specific_key], specific_value)
```

vyasr (approved):
One small suggestion, otherwise LGTM.

PR #64, opened 2024-01-02
refactor: remove default in delete_existing_files()

jameslamb:
`delete_existing_files()` is used to clean up previous work done by `dependency-file-generator`.

This PR proposes removing the default path that function uses when looking for files to delete.

### How is this safe?

That function is only used in one place, and a value is explicitly passed in that place.

```shell
git grep delete_existing_files
```

https://github.com/rapidsai/dependency-file-generator/blob/36a40613235089b4de271367ccef3674678c1555/src/rapids_dependency_file_generator/cli.py#L103

### Benefits of this change

Makes the flow of configuration stricter, reduces the risk of "forgot to pass through this configuration and silently fell back to a default" types of bugs.

### Notes for Reviewers

This isn't in response to any particular issue.

I'm just reading through the library to understand it better, and noticed this.


msarahan (approved):
At some point it might be nice to use pathlib instead of str, but that's much larger scope than this. This is a good improvement for now.

ajschmidt8 (approved):

PR #67, opened 2024-02-22
feat: Add support for prepending conda channels via the CLI.

bdice:
This adds an option `--prepend-channels "my_channel;my_other_channel"` to the `rapids-dependency-file-generator` CLI. This will allow us to use `rapids-dependency-file-generator` with local channels containing PR artifacts fetched in CI workflows. If we combine this feature with some small changes in `dependencies.yaml`, we will be able to generate the entire CI environment in one pass rather than generating an environment with test dependencies and then installing the packages (e.g. `libcuml` and `libcuml-tests`) in a separate step.

This option is modeled after `conda config --prepend channels new_channel` ([reference](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html)).

See original proposal in https://github.com/rapidsai/cuml/pull/5781#issuecomment-1958498274 (this PR is slightly improved from that proposal).

This is a partial solution for https://github.com/rapidsai/build-planning/issues/22.


bdice:
To solve https://github.com/rapidsai/build-planning/issues/22, I'm thinking we'll do this kind of thing in each RAPIDS repo CI test script:

```bash
rapids-logger "Downloading artifacts from previous jobs"
CPP_CHANNEL=$(rapids-download-conda-from-s3 cpp)
PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)

  --output conda \
  --file_key test_python \
  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" \
  --prepend-channels "${CPP_CHANNEL};${PYTHON_CHANNEL}" | tee env.yaml

rapids-mamba-retry env create --force -f env.yaml -n test
```

We would modify the `test_python` file key to include a dependency list containing `libcuml` and `cuml`. Then this environment would be equivalent to the `conda-merge` output I was using in https://github.com/rapidsai/cuml/pull/5781.

ajschmidt8 (approved):
LGTM.

PR #69, opened 2024-03-08
feat: Relax pyproject table/key schema

vyasr:
Currently dfg only allows dependencies in pyproject.toml files to be written to a strictly proscribed set of places. This list prohibits more general use cases like the one I intend to employ with https://github.com/vyasr/rapids_builder/.


bdice (approved):

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -165,9 +165,29 @@ def make_dependency_file(
                 table[section] = tomlkit.table()
                 table = table[section]
 
-        key = extras.get(
-            "key", "requires" if extras["table"] == "build-system" else "dependencies"
-        )
+        if extras["table"] == "build-system":
+            key = "requires"
+            if key in extras:
```
KyleFromNVIDIA, line 170: Should this have been:

```suggestion
            if "key" in extras:
````

?

vyasr: Yes, this is definitely a typo. Feel free to fix in your next PR. The same issue also exists on line 176 here.

PR #70, opened 2024-03-19
feat: Add matrix support to pyproject.toml

KyleFromNVIDIA:
Co-authored-by: @trxcllnt


trxcllnt (approved):

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -188,8 +161,33 @@ def make_dependency_file(
                     "'table' is not one of 'build-system' or 'project'."
                 )
 
+        # This file type needs to be modified in place instead of built from scratch.
```
vyasr, line 164: Did you move all the code just so we can error earlier in the above cases? AFAICT nothing actually changed here other than that you call tomlkit.dumps on the end result, which I guess is to isolate the divergence in file types here from where it used to be in `make_dependency_files`.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -138,43 +138,16 @@ def make_dependency_file(
...
-            file_contents = tomlkit.load(f)
-
-        toml_deps = tomlkit.array()
-        for dep in dependencies:
-            toml_deps.add_line(dep)
-        toml_deps.add_line(indent="")
-        toml_deps.comment(
-            f"This list was generated by `{cli_name}`. To make changes, edit "
-            f"{relative_path_to_config_file} and run `{cli_name}`."
-        )
```
vyasr, line 152: We lost the comment.

vyasr (changes requested):
I think we need to add some testing/error checking for some additional cases.
- What happens if you have a dependencies.yaml where a matrix is specified in the `files` section for a file where the output is pyproject? Right now I think it'll just overwrite in sequence and you'll end up with the last matrix entry, but I think that case should probably still error. Perhaps more precisely, we should only allow a single matrix entry. If the user wants to use a matrixed dependency list with a pyproject file key, that is OK but they either have to use no matrix and fall back to defaults, or they need to specify a single matrix element to build with
- We should add a test for matrix fallback behavior with pyproject outputs. I think everything there should work fine since the logic for selecting dependencies from dependency lists is handled independently of any file type-specific code path, but we should verify that.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
KyleFromNVIDIA: That was carried over from #48. I'll put it back.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
KyleFromNVIDIA: > Did you move all the code just so we can error earlier in the above cases?

Yes.

> you call tomlkit.dumps on the end result, which I guess is to isolate the divergence in file types here from where it used to be in `make_dependency_files`.

Correct.

KyleFromNVIDIA:
> Perhaps more precisely, we should only allow a single matrix entry.

Done

vyasr (approved):

PR #71, opened 2024-03-22
feat: Update CLI arguments

KyleFromNVIDIA:


bdice:
We use `--file_key` a lot in RAPIDS. https://github.com/search?q=org%3Arapidsai+%22--file_key%22&type=code

When this PR is merged, it would create new CI images that expect the new name and thus would break CI for all repos using `--file_key`. Can we have a temporary way to enable both, and port repos to the new keyword before removing the old one?

vyasr (changes requested):
I agree with Bradley, can we make this a non-breaking PR that just adds the new args (we can error if old and new args are mixed)? Then we can remove the old args in a follow-up after a deprecation period where we have time to clean up everywhere using them.

On a related note, this repo uses [semantic-release](https://github.com/semantic-release/semantic-release) to manage versioning. If your PR is breaking, you should [title it appropriately](https://github.com/semantic-release/semantic-release?tab=readme-ov-file#commit-message-format) so that the automated release engine knows that the next release should be a major release. For major releases it looks like the expected title prefix is "BREAKING CHANGE: ". Note that it mentions commit messages, but those are effectively the title because we squash.

KyleFromNVIDIA:
I've restored both `--file_key` and `--prepend-channels`.

src/rapids_dependency_file_generator/cli.py
```diff
@@ -35,10 +37,17 @@ def validate_args(argv):
 
     codependent_args = parser.add_argument_group("optional, but codependent")
     codependent_args.add_argument(
-        "--file_key",
+        "-f",
```
bdice, line 40: Can we remove the short form `-f`? I like forcing callers to spell out `--file-key` to make it clear for readers who aren't familiar with DFG. This would also avoid confusion with `-f` meaning `--force` (this is a common point of confusion in conda which mixes `--file` and `--force`, both abbreviated as `-f` depending on the command).
```suggestion
```

src/rapids_dependency_file_generator/cli.py
```diff
@@ -35,10 +37,17 @@ def validate_args(argv):
...
+        "--file-key",
         help="The file key from `dependencies.yaml` to generate",
     )
     codependent_args.add_argument(
+        "--file_key",
+        dest="file_key_deprecated",
+        help="Deprecated alias for --file-key",
```
bdice, line 47: Do you want to spell this deprecation the same way as below?
```suggestion
        help="DEPRECATED: Use --file-key instead.",
```

src/rapids_dependency_file_generator/cli.py
```diff
@@ -35,10 +37,17 @@ def validate_args(argv):
...
+    )
+    codependent_args.add_argument(
+        "-o",
```
bdice, line 50: Same as above, remove the short flags for consistency / explicitness.
```suggestion
```

src/rapids_dependency_file_generator/cli.py
```diff
@@ -51,35 +60,62 @@ def validate_args(argv):
         ],
     )
     codependent_args.add_argument(
+        "-m",
```
bdice, line 63: Same as above, remove the short flags for consistency / explicitness.
```suggestion
```

src/rapids_dependency_file_generator/cli.py
```diff
@@ -51,35 +60,62 @@ def validate_args(argv):
...
             '`--prepend-channels "my_channel;my_other_channel"`. This option is '
-            f"only valid with --output {OutputTypes.CONDA} or no --output"
+            f"only valid with --output {OutputTypes.CONDA} or no --output. "
+            "DEPRECATED: Use --prepend-channel instead."
         ),
     )
 
     args = parser.parse_args(argv)
+
+    if args.file_key_deprecated:
```
bdice, line 96: Don't let users specify both flags.
```suggestion
    if args.file_key_deprecated:
        if args.file_key:
            raise ValueError("The --file_key (deprecated) and --file-key arguments cannot be specified together.")
```

ajschmidt8 (approved):

src/rapids_dependency_file_generator/cli.py
```diff
@@ -51,35 +60,62 @@ def validate_args(argv):
...
     dependent_arg_keys = ["file_key", "output", "matrix"]
     dependent_arg_values = [getattr(args, key) is None for key in dependent_arg_keys]
     if any(dependent_arg_values) and not all(dependent_arg_values):
         raise ValueError(
             "The following arguments must be used together:"
-            + "".join([f"\n  --{x}" for x in dependent_arg_keys])
+            + "".join([f"\n  {x}" for x in ["--file-key", "--output", "--matrix"]])
         )
 
+    if args.prepend_channels_deprecated:
```
bdice, line 111: Raise if both are used.
```suggestion
    if args.prepend_channels_deprecated:
        if args.prepend_channels:
            raise ValueError("The --prepend-channels (deprecated) and --prepend-channel arguments cannot be specified together.")
```

vyasr (approved):
Happy with this assuming Bradley's concerns are addressed.

README.md
```diff
@@ -329,12 +329,12 @@ mamba activate "$ENV_NAME"
...
+The `--file-key` argument is passed the `test` key name from the `files` configuration. Additional flags are used to generate a single dependency file. When the CLI is used in this fashion, it will print to `stdout` instead of writing the resulting contents to the filesystem.
 
-The `--file_key`, `--output`, and `--matrix` flags must be used together. `--matrix` may be an empty string if the file that should be generated does not depend on any specific matrix variations.
+The `--file-key`, `--output`, and `--matrix` flags must be used together. `--matrix` may be an empty string if the file that should be generated does not depend on any specific matrix variations.
 
-The `--prepend-channels` argument accepts a semi-colon separated list of additional channels to use, like `rapids-dependency-file-generator --prepend-channels "my_channel;my_other_channel"`.
-If both `--output` and `--prepend-channels` are provided, the output format must be conda.
+The `--prepend-channel` argument accepts additional channels to use, like `rapids-dependency-file-generator --prepend-channel "my_channel" --prepend-channel "my_other_channel"`.
```
bdice, line 336: We probably don't need to quote this, now that the semicolon is not in the string. I'm fine either way if you like the quotes.
```suggestion
The `--prepend-channel` argument accepts additional channels to use, like `rapids-dependency-file-generator --prepend-channel my_channel --prepend-channel my_other_channel`.
```

bdice (approved):

README.md
KyleFromNVIDIA: Removed.

PR #74, opened 2024-04-03
chore(refactor): Add Config class

KyleFromNVIDIA:


src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -79,12 +74,12 @@ def grid(gridspec):
...
     Parameters
     ----------
-    gridspec : dict
+    gridspec : dict[str, list[str]]
         A mapping from parameter names to lists of parameter values.
 
     Yields
     ------
-    Iterable[dict]
+    Generator[dict[str, str]]
```
bdice, line 82: I think the "Yields" section implies that this is a generator already. This yields dictionaries, but the docs imply that it yields generators of dictionaries.

https://numpydoc.readthedocs.io/en/latest/format.html#yields

As a note, the return type above of `-> Generator[...]` should be fine as-is.
```suggestion
    dict[str, str]
```

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -3,33 +3,26 @@
 import os
 import textwrap
 from collections import defaultdict
+from collections.abc import Generator
```
bdice, line 6: This should probably be `typing.Generator` rather than an ABC. It's just used for a type hint.

https://docs.python.org/3/library/typing.html#typing.Generator

```suggestion
from typing import Generator
```

bdice (commented):
A couple minor comments. Overall this is really great work, I'm super happy with it.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
KyleFromNVIDIA: From https://docs.python.org/3/library/typing.html#typing.Generator:

> Deprecated alias to [`collections.abc.Generator`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Generator).

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
KyleFromNVIDIA: Updated

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
bdice: Wow, if only I could read the link that I shared. :) Resolving this.

bdice (approved):

ajschmidt8 (approved):
:fire: :fire:

tests/test_config.py
```diff
@@ -0,0 +1,417 @@
...
+            config.parse_config(input, path)
+    else:
+        assert config.parse_config(input, path) == output
+
+
+@pytest.mark.parametrize(
+    ["input", "path", "output"],
+    [
+        (
+            """\
```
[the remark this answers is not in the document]
KyleFromNVIDIA: OMG THANK YOU. I'll be using that quite a bit.

PR #77, opened 2024-04-11
chore(doc): Add Sphinx build

KyleFromNVIDIA:


src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
```diff
@@ -329,32 +333,33 @@ def make_dependency_files(
...
-        Whether the output should be written to stdout. If False, it will be
-        written to a file computed based on the output file type and
-        config_file_path.
-
-    Raises
-    -------
-    ValueError
-        If the file is malformed. There are numerous different error cases
-        which are described by the error messages.
+    :param parsed_config: The parsed dependencies.yaml config file.
```
bdice, line 339: This changed away from NumPy style. We use NumPy style docstrings in most of RAPIDS. I know it’s a lot of work to switch but we should just be consistent and use NumPy style.

bdice (changes requested):
Please use NumPy style docstrings for consistency with most of RAPIDS.

src/rapids_dependency_file_generator/rapids_dependency_file_generator.py
vyasr: If there's a lot you can try using `pyment` or `docconvert` to automate the conversion.

src/rapids_dependency_file_generator/__init__.py
```diff
@@ -1,5 +1,10 @@
+from . import config, rapids_dependency_file_generator
 from ._version import __version__
+from .config import *  # noqa: F401,F403
```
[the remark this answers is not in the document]
KyleFromNVIDIA: The idea is that each package specifies what it wants to export with `__all__`. IMO the top-level module should not be responsible for this.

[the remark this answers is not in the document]
bdice: It's fine, I guess. I just like to adhere to https://www.flake8rules.com/rules/F403.html. I tend to recommend against `import *`.

bdice (approved):

docs/source/_templates/class.rst
```diff
@@ -0,0 +1,32 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+   :members:
+   :show-inheritance:
+   :inherited-members:
```
vyasr, line 8: It looks like these directives are the main addition being made to the default template here. Do you really think it's worthwhile? I'm weakly opposed since so few people actually know enough about how autosummary works to know what this is doing, but I'm OK with it if you really think it's worth it.

docs/source/_templates/module.rst
vyasr, line 1: Hmm OK I guess you're also including a template here to handle adding things to toctrees correctly as well as use the class template above. Again, I'm not sure it's worthwhile but I'll defer to your judgment.

docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
```
vyasr, line 6: We don't have to do it in this PR, but I would recommend that in the follow-up where we enable ruff docstring linting we also add `-n -v -W --keep-going`. The `-n` and `-W` make the doc build throw a lot more errors on things like missing docstrings or incorrect crosslinks caused by typos etc, while the `--keep-going` lets you see all the errors rather than stopping on the first one (`-v` is verbose as usual). I'd recommend trying it in this PR and seeing how extensive the errors are, and if there are a lot we can defer to a future PR. It seems like you already added a bunch of missing docstrings, which I imagine was to silence these kinds of warnings.

docs/source/conf.py
vyasr, line 1: You probably want to set up an [intersphinx mapping](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping) in this file. We expose a couple of objects in public APIs that ought to be linked to, for instance `validate_dependencies` can throw a `jsonschema` exception.

vyasr (changes requested):
We don't have to do it in this PR, but once this is merged can we replace all the current linters possible with ruff? When we do, we can also enable [ruff's rules for docstring linting](https://docs.astral.sh/ruff/rules/#pydocstyle-d) to help ensure that these docstrings stay in good shape (complementary to building with warnings as errors as I suggested in an inline comment). It might be worth trying to just add ruff in this PR for the purpose of docstring linting and seeing how extensive the changes are, in case they're not too bad.

Most of the changes that I suggested are optional, but it's worth trying them out before we merge in case they're easy and we can just knock them out here.

docs/source/_templates/class.rst
bdice: I feel the same as Vyas -- but I didn't think this was worth changing. If you're happy, I think it's good enough, and we should expend effort on the next steps.

docs/Makefile
bdice: Good idea!

docs/source/_templates/module.rst
KyleFromNVIDIA: I got these templates from StackOverflow:

https://stackoverflow.com/questions/2701998/sphinx-autodoc-is-not-automatic-enough/62613202#62613202

Linked from this issue:

https://github.com/sphinx-doc/sphinx/issues/7912

pyproject.toml
```diff
@@ -36,3 +36,9 @@ version = {attr = "rapids_dependency_file_generator._version.__version__"}
 
 [tool.isort]
 profile = "black"
+
+[tool.ruff.lint]
+select = ["F", "I", "D"]
```
bdice, line 41:
```suggestion
select = ["E", "F", "W", "I", "D"]
```

vyasr (approved):
Approved, assuming you add the remaining ruff rules we need. Thanks!

Issue #76, opened 2024-04-05
Make the boundary between public and private in the API clearer


jameslamb:
## Details

The ongoing work on `rapids-build-backend` relies on using `rapids-dependency-file-generator` as an importable module, not a CLI, like this

```python
from rapids_dependency_file_generator.cli import generate_matrix
from rapids_dependency_file_generator.constants import default_pyproject_dir
from rapids_dependency_file_generator.rapids_dependency_file_generator import (
    get_requested_output_types,
    make_dependency_files,
)
```

ref: https://github.com/rapidsai/rapids-build-backend/pull/17/files#r1542068869

Given that, I think some work should be done to make the boundary between public and private parts of the API here clearer.

## Approach

* use `__all__` entries in every module of the library
   - to limit which things should be imported with `*` imports
* use an `__all__` entry in the top-level `__init__.py`
    - to define what is importable directly from the `rapids_dependency_file_generator` module without needing to know about submodules
    - to limit what is imported by `from rapids_dependency_file_generator import *`
* prefix all internal-only functions, classes, methods, and other objects with `_`

## Benefits of this work

Clarifies the boundary between public and private, reducing the need for tight coordination between `rapids-build-backend` and `rapids-dependency-file-generator` for some types of changes.

Makes it easier for tools to help enforce those boundaries:

* `sphinx` autodoc will by default only auto-generate documentation for entries in `__all__` ([docs](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-members))
* some linters will complain about uses of imports prefixed with `_` (e.g. [`ruff`'s pylint `import-private-name` check](https://docs.astral.sh/ruff/rules/import-private-name/)))

## Other notes

Consider the following.

<details><summary>only '__version__' is importable from the top-level namespace today (click me)</summary>

```shell
pip install .
python -c "from rapids_dependency_file_generator import *; print(dir())"
```

```text
['__annotations__', '__builtins__', '__doc__', '__loader__', '__name__', '__package__', '__spec__', '__version__']
```

Only `__version__` is exported from the top-level module

https://github.com/rapidsai/dependency-file-generator/blob/795c0c3afb55f6a11d36de7fa9b8f74001891327/src/rapids_dependency_file_generator/__init__.py#L3-L5

That means that things like `rapids-build-backend` that want to import other things have imported coupled to the exact file layout in this project, like this.

```python
from rapids_dependency_file_generator.rapids_dependency_file_generator import (
[...]

PR #78, opened 2024-04-19
chore(ci): Build docs

KyleFromNVIDIA:


bdice (changes requested):
Can we make the docs build into a separate CI job, and end with calling `rapids-upload-docs`? Try to mirror this script: https://github.com/rapidsai/cudf/blob/branch-24.06/ci/build_docs.sh

Then we'll be able to preview the docs as CI artifacts.

ajschmidt8:
> Can we make the docs build into a separate CI job, and end with calling `rapids-upload-docs`? Try to mirror this script: https://github.com/rapidsai/cudf/blob/branch-24.06/ci/build_docs.sh
> 
> Then we'll be able to preview the docs as CI artifacts.

There's a few things to be aware of here.

`rapids-upload-docs` requires AWS credentials to upload to S3.

We use this OIDC action to obtain temporary AWS credentials https://github.com/aws-actions/configure-aws-credentials/.

That action requires that the `id-token` permission on the calling workflow be set to `write`.

However, when using the `pull_request` event to trigger workflows from forks, that permission can only ever have a maximum permission of `read`: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token.

Therefore if you want to set `id-token: write`, you'll need to either move away from using forked PRs (e.g. open PRs from branches within the source repository) or switch to `copy-pr-bot`, which is how we get away with it in other RAPIDS libraries.

.github/workflows/build-test.yaml
```diff
@@ -23,6 +23,13 @@ jobs:
         with:
           name: conda_artifacts
           path: ${{ env.OUTPUT_DIR }}
+  docs-build:
+    if: github.ref_type == 'branch'
+    needs: conda
+    runs-on: ubuntu-latest
+    container: rapidsai/ci-conda:latest
+    steps:
+      - run: ./ci/build_docs.sh
```
ajschmidt8, line 32: you'll need to use [actions/checkout](https://github.com/actions/checkout) first

ci/build_docs.sh
```diff
@@ -0,0 +1,41 @@
...
+  --output conda \
+  --file_key docs | tee "${ENV_YAML_DIR}/env.yaml"
+
+rapids-mamba-retry env create --yes -n docs
+conda activate docs
+
+rapids-print-env
+
+rapids-logger "Downloading artifacts from previous jobs"
+PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)
```
ajschmidt8, line 21: I don't think this command will work as expected on this PR.

The previous jobs in the workflow don't upload artifacts to S3, they use GitHub's native artifact storage.

jameslamb:
@KyleFromNVIDIA are you planning to return to this? It's all that's keeping #76 open: https://github.com/rapidsai/dependency-file-generator/issues/76#issuecomment-2064575569

If you do in the near future, just noting you might need to put an upper bound on `sphinx`: https://github.com/rapidsai/build-planning/issues/155

KyleFromNVIDIA:
I hope to get back to this some time in the next few days.

Issue #87, opened 2024-05-13
Enforce `mypy` checks


jameslamb:
## Description

The Python code in this project should be checked with `mypy`. It isn't currently checked by any static type-checker.

## Benefits of this work

* validates the type hints, improving their usefulness as documentation
* increases release confidence via improving the chance of some types of bugs being caught during development
* reduces the effort required to review PRs

## Acceptance Criteria

* `mypy` is configured to run over this project's Python code via `pre-commit`

## Approach

Keep as much configuration as possible in `pyproject.toml`, the rest in `pre-commit`'s config. e.g.

https://github.com/rapidsai/cudf/blob/149253b2e9f3801fdcc88c17e31a25788fe6381a/pyproject.toml#L3

https://github.com/rapidsai/cudf/blob/149253b2e9f3801fdcc88c17e31a25788fe6381a/.pre-commit-config.yaml#L32

## Notes

As of this writing, the latest version of `mypy` (v1.10)`, run like this in a Python 3.10 environment ...

```shell
mypy \
    --ignore-missing-imports \
    --explicit-package-bases \
    ./src
```

... yields the following

<details><summary>Found 26 errors in 2 files (checked 7 source files) (click me)</summary>

```text
src/rapids_dependency_file_generator/_config.py:7: error: Library stubs not installed for "yaml"  [import-untyped]
src/rapids_dependency_file_generator/_config.py:7: note: Hint: "python3 -m pip install types-PyYAML"
src/rapids_dependency_file_generator/_config.py:7: note: (or run "mypy --install-types" to install all missing stub packages)
src/rapids_dependency_file_generator/_config.py:7: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
src/rapids_dependency_file_generator/_config.py:171: error: Argument 1 to "_parse_outputs" has incompatible type "object"; expected "str | list[str]"  [arg-type]
src/rapids_dependency_file_generator/_config.py:173: error: No overload variant of "list" matches argument type "object"  [call-overload]
src/rapids_dependency_file_generator/_config.py:173: note: Possible overload variants:
src/rapids_dependency_file_generator/_config.py:173: note:     def [_T] __init__(self) -> list[_T]
src/rapids_dependency_file_generator/_config.py:173: note:     def [_T] __init__(self, Iterable[_T], /) -> list[_T]
src/rapids_dependency_file_generator/_config.py:174: error: "object" has no attribute "items"  [attr-defined]
[...]

PR #90, opened 2024-05-13
fix: fix return type hint for dedupe()

jameslamb:
Contributes to #87.

`rapids-dependency-file-generator` handles deduplicating the sets of dependencies provided to it, including `pip: ` lists for conda environments.

This PR modifies the logic the function does that in the following ways:

* fixes return type hint
* refactors it so that:
   - only a single pass is done over the inputs to split requirements in to pip vs non-pip
   - assignments are done in a way that's a bit easier for a type checker to understand

This fixes these errors found by `mypy`:

```text
src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py:65: error: Argument 1 to "append" of "list" has incompatible type "dict[str, list[str]]"; expected "str"  [arg-type]
src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py:66: error: Incompatible return value type (got "list[str]", expected "list[str | dict[str, str]]")  [return-value]
src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py:66: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py:66: note: Consider using "Sequence" instead, which is covariant
src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py:66: note: Perhaps you need a type annotation for "deduped"? Suggestion: "list[str | dict[str, str]]"
src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py:425: error: Argument "dependencies" to "make_dependency_file" has incompatible type "list[str | dict[str, str]]"; expected "list[str | dict[str, list[str]]]"  [arg-type]
```

## How I tested this

Confirmed that this behavior is covered by existing unit tests (and added one more condition there for completeness).

Ran the following to confirm that some of the `mypy` issues had been resolved:

```shell
mypy \
    --ignore-missing-imports \
    --explicit-package-bases \
    ./src
```

### DIFF
--- a/src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py
+++ b/src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py
@@ -3,7 +3,6 @@ import itertools
 import os
 import textwrap
 import typing
-from collections import defaultdict
 from collections.abc import Generator
 
 import tomlkit
@@ -42,7 +41,7 @@ def delete_existing_files(root: os.PathLike) -> None:
 
 def dedupe(
     dependencies: list[typing.Union[str, _config.PipRequirements]],
-) -> list[typing.Union[str, dict[str, str]]]:
+) -> list[typing.Union[str, dict[str, list[str]]]]:
     """Generate the unique set of dependencies contained in a dependency list.
 
     Parameters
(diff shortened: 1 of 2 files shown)


src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py
```diff
@@ -52,18 +51,21 @@ def dedupe(
...
+    for dep in dependencies:
+        if isinstance(dep, str):
+            string_deps.add(dep)
+        elif isinstance(dep, _config.PipRequirements):
+            pip_deps.update(dep.pip)
+
+    if pip_deps:
+        return [*sorted(string_deps), {"pip": sorted(pip_deps)}]
+    else:
+        return sorted(string_deps)  # type: ignore[return-value]
```
jameslamb, line 68: This `#type: ignore` is here to avoid this:

```text
src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py:68: error: Incompatible return value type (got "list[str]", expected "list[str | dict[str, list[str]]]")  [return-value]
src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py:68: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py:68: note: Consider using "Sequence" instead, which is covariant
```

`mypy` is just really unhappy with the idea of `list[str]` being a strict subset of `list[str | dict[str, list[str]]]`.

KyleFromNVIDIA: Why not use `Sequence` instead, as the warning suggested?

jameslamb: I'd tried a couple of variations using `Sequence` and then abandoned it, because doing that means you can't use methods that are common to all `Sequence` types.

e.g. I'd stuff like tried this:

```python
out: Sequence[Union[str, dict[str, list[str]]]] = sorted(string_deps)
if pip_deps:
    out.append({"pip": sorted(pip_deps)})](out: typing.Sequence[typing.Union[str, dict[str, list[str]]]] = sorted(string_deps))
```

That generated errors like this:

```text
error: "Sequence[str | dict[str, list[str]]]" has no attribute "append"
```

HOWEVER.... I never revisited that after switching to this multiple-return-statements pattern! (which avoids the `.append()`)

I just pushed https://github.com/rapidsai/dependency-file-generator/pull/90/commits/ddb8c7f53e535350369f5ba587c9b45edff23a1b switching to `Sequence` and removing this `# type: ignore`, thanks for the nudge.

ajschmidt8 (approved):

PR #100, opened 2024-06-17
feat(cli): remove --file_key and --prepend-channels arguments

### LINKED ISSUE #89: Remove "--file_key" and "--prepend_channels"
## Description

In #71, the arguments `--file_key` and `--prepend-channels` were deprecated.

Uses of them across RAPIDS should be removed, and then they and the corresponding deprecation warnings here should be removed.

## Benefits of this work

* simplifies `rapids-dependency-file-generator`
* removes a source of noise in CI logs across RAPIDS projects

## Acceptance Criteria

* 0 non-archived projects in the `rapidsai` org use `--file_key` or `--prepend-channels` in their calls to `rapids-dependency-file-generator`
* arguments `--file_key` and `--prepend-channels` do not exist in `rapids-dependency-file-generator`

## Approach

Across the `rapidsai` GitHub org:

* replace `--file_key` with `--file-key` *(note the `-`)*
* replace `--prepend-channels` with one or more `--prepend-channel` 

For example this:

```shell
  --output conda \
  --file_key test_cpp \
  --matrix "cuda=12.2" \
  --prepend-channels "${CPP_CHANNEL};pytorch"
```

Should become this:

```shell
  --output conda \
  --file-key test_cpp \
  --matrix "cuda=12.2" \
  --prepend-channel "${CPP_CHANNEL}" \
  --prepend-channel "pytorch"
```

## Notes

N/A

jameslamb:
Fixes #89

Removes the arguments `--file_key` and `--prepend-channels`.

Those have been deprecated since #71.

## Notes for Reviewers

I've removed all remaining uses of them across the `rapidsai`, `nv-morpheus`, and `NVIDIA` GitHub organizations. See the PRs linked to https://github.com/rapidsai/dependency-file-generator/issues/89.

### Shouldn't this be a breaking change (and a 1.0 to 2.0 bump)?

This is a breaking change to the CLI but not to the library. And because so many uses of the deprecated pattern have already been removed across NVIDIA repos, there should be minimal cases where even the CLI breaks.

I'm proposing committing with a message like `feat:...` so that `semantic-release` will publish this as v1.14.0, not v2.0.0.

I'm proposing that this cause a minor version bump so that we don't have to go update `rapids-build-backend`'s constraint on `rapids-dependency-file-generator`:

https://github.com/rapidsai/rapids-build-backend/blob/4e6969363bbddf8b4e94cf354077319337b5c432/pyproject.toml#L17


KyleFromNVIDIA (approved):
I am thinking about the slim possibility that someone outside of RAPIDS is using this project and would have their project broken by this removal. Weighing that against:

> I'm proposing that this cause a minor version bump so that we don't have to go update `rapids-build-backend`'s constraint on `rapids-dependency-file-generator`:

Updating the constraint really doesn't seem like a big deal to me, but I do understand that there is a very slim chance of this change causing a problem for anyone outside of RAPIDS. I am slightly in favor of bumping to a major version, but I won't push the matter.

jameslamb:
Thanks for considering it.

> *Updating the constraint really doesn't seem like a big deal to me*

I personally think it's not worth it the effort. But I'm happy to let @bdice be the tie-breaker on whether this is `1.4.0` or `2.0.0`.

If it's `2.0.0`, we'll have to follow it up with these additional changes:

* update `rapids-build-backend`'s constraint ([rapidsai/rapids-build-backend code link](https://github.com/rapidsai/rapids-build-backend/blob/4e6969363bbddf8b4e94cf354077319337b5c432/pyproject.toml#L17)) and do a new release of `rapids-build-backend`
* update constraints in CI images ([rapidsai/ci-imgs code link](https://github.com/rapidsai/ci-imgs/blob/d1b97efbbfc0796e3579a86b140b306f46f5cb72/ci-conda.Dockerfile#L134))
  - *I just found this one this morning*

KyleFromNVIDIA:
Weren't we also considering adding the `--stdout` argument from #48 before a 2.0 release, since that would be a breaking change? Cc: @trxcllnt

bdice:
I do not know of any consumers of rapids-dependency-file-generator outside of RAPIDS "core" + a few other NVIDIA repositories that are RAPIDS-adjacent. However it looks like we have several private repos where breaking the CLI without a major bump might break CI. I like being expedient where possible but I would check over this list and make sure you have investigated the possibility of breaking CI for these repos: https://github.com/search?q=org%3Arapidsai+--file_key&type=code

jameslamb:
Thanks.

> *it looks like we have several private repos where breaking the CLI without a major bump might break CI*

If you add an `AND NOT is:archived` filter ([this search link](https://github.com/search?q=org%3Arapidsai+--file_key+AND+NOT+is%3Aarchived&type=code)), the only remaining private repo that would be broken by this is the `*-cuml` one. @dantegd told me that's not being actively maintained right now, and that it's safe to ignore for RAPIDS-wide changes until that changes. I can share a link to our conversation with you internally.

It **does** look like I missed one public repo... `cugraph-gnn`, probably because that repo is so new. I've put up https://github.com/rapidsai/cugraph-gnn/pull/5.

> *I would check over this list and make sure you have investigated the possibility of breaking CI for these repos*

I did that (with the `AND NOT is:archived`) filter for the `rapidsai`, `NVIDIA`, and `nv-morpheus` organizations. I'm confident that the only remaining place in those GitHub organizations that'd be broken by publishing this as 1.4.0 is `cugraph-gnn`.

If you still feel after reading that that it'd be better to make this 2.0, I'll change the title of this so that it'd be published as 2.0 and then go make the `ci-imgs` and `rapids-build-backend` changes after this is published.

> *Weren't we also considering adding the `--stdout` argument from https://github.com/rapidsai/dependency-file-generator/pull/48 before a 2.0 release*

Can we please not couple that to this change, especially since (as far as I know) it's not currently being actively worked on?

PR #102, opened 2024-07-23
test: add tests on handling of duplicate matrix selectors

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/31.

Over in https://github.com/rapidsai/devcontainers/pull/365, I'm proposing a change that depends on how `rapids-dependency-file-generator` handles duplicates in `--matrix` selectors.

See https://github.com/rapidsai/devcontainers/pull/365#discussion_r1687121294.

This proposes adding tests covering that behavior, to reduce the risk of accidentally removing that support in future refactorings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,8 @@ The `--file-key` argument is passed the `test` key name from the `files` configu
 
 The `--file-key`, `--output`, and `--matrix` flags must be used together. `--matrix` may be an empty string if the file that should be generated does not depend on any specific matrix variations.
 
+Where multiple values for the same key are passed to `--matrix`, e.g. `cuda_suffixed=true;cuda_suffixed=false`, only the last value will be used.
+
 The `--prepend-channel` argument accepts additional channels to use, like `rapids-dependency-file-generator --prepend-channel my_channel --prepend-channel my_other_channel`.
 If both `--output` and `--prepend-channel` are provided, the output format must be conda.
 Prepending channels can be useful for adding local channels with packages to be tested in CI workflows.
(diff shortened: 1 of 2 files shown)


KyleFromNVIDIA (changes requested):
This is fine, but if it was me, I would have modified https://github.com/rapidsai/devcontainers/pull/365#discussion_r1687121294 to detect if the user is overriding `cuda_suffixed` to avoid passing more than one value.

Let's explicitly document this behavior too.

jameslamb:
> *This is fine, but if it was me, I would have modified https://github.com/rapidsai/devcontainers/pull/365#discussion_r1687121294 to detect if the user is overriding cuda_suffixed to avoid passing more than one value.*

The simplest implementation I can think of involves tracking the defaults as a bash associative array, splitting each `--matrix-entry` value by `=` into `(key, value)`, and then indexing into that associative array with them.

That was more complexity than I was comfortable with for this purpose, but I'm willing to try it. Let's please not hold up https://github.com/rapidsai/devcontainers/pull/365 or this PR for that though. I'd like to try to get https://github.com/rapidsai/devcontainers/pull/365 and the corresponding `dependencies.yaml` changes out before code freeze for 24.08 (end of day tomorrow). 

> *Let's explicitly document this behavior too.*

Document what, where? I don't understand what you're asking for, sorry.

KyleFromNVIDIA:
> Document what, where? I don't understand what you're asking for, sorry.

Document the fact that the last duplicate is the one that DFG actually uses, which is what this PR is all about.

KyleFromNVIDIA (approved):

KyleFromNVIDIA:
The README, because that's what the end user is going to see.

jameslamb:
Gotchu! Ok yeah no problem, I agree that's a good addition.

Thanks for the quick review, always appreciated!

PR #106, opened 2024-08-28
chore(refactor): update pre-commit hook verions, pyupgrade target Python version

jameslamb:
This project supports `python>=3.9`

https://github.com/rapidsai/dependency-file-generator/blob/4e32ded33b3df70e88cff06bac59488fca74ae7f/pyproject.toml#L21

But is telling `pyupgrade` it supports `python>=3.8`.

https://github.com/rapidsai/dependency-file-generator/blob/4e32ded33b3df70e88cff06bac59488fca74ae7f/.pre-commit-config.yaml#L22-L25

This fixes that, and updates all the other hook versions (`pre-commit autoupdate`) since it's been a while.

### Shouldn't we upgrade the floor to `requires-python = >= 3.10`?

I don't think so. That's being done across RAPIDS libraries in https://github.com/rapidsai/build-planning/issues/88 right now, but `rapids-dependency-file-generator` is used in many places outside of RAPIDS (see everything linked to https://github.com/rapidsai/dependency-file-generator/issues/89, for example).

For tools like this, I think we should generally support a wider range of Python versions than RAPIDS itself does.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,20 +11,20 @@ repos:
       - id: debug-statements
       - id: requirements-txt-fixer
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: 'v1.10.0'
+    rev: 'v1.11.2'
     hooks:
       - id: mypy
         additional_dependencies: [types-PyYAML]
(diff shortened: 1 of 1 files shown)


jameslamb:
CI here will be fixed by #107

bdice (approved):

PR #110, opened 2024-09-18
feat: Correct behavior of matrix fallback priority, error on duplicate matrix selectors.

bdice:
The README says,

> Every `matrices` list must have a match for a given input matrix (only the first matching matrix in the list of `matrices` will be used).

However, it seems that the implemented behavior does not align with this. Only the fallback matrix worked in this way. This PR fixes the implementation to match the README (and our expectations).

We also decided it would be nice to fail if there are duplicate matrix selectors, since only the first one would be usable.

I added tests covered by both of these features/fixes.


src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py
```diff
@@ -403,18 +416,12 @@ def make_dependency_files(
...
-                                    raise ValueError(f"Found multiple matches for matrix {matrix_combo}")
-                                found = True
                                 # A package list may be empty as a way to
                                 # indicate that for some matrix elements no
                                 # packages should be installed.
                                 dependencies.extend(specific_matrices_entry.packages or [])
-
-                        if not found:
+                                break
+                        else:
```
bdice, line 424: On the off chance that this is new to you: Python has a `for/else` construct that can be used to simplify things where you would otherwise need a loop with a "found" variable. It doesn't come up often in Python usage, but it's nice when it fits the problem cleanly. I was able to eliminate the `found` variable here with that approach.

jameslamb: Thanks for pointing this out, this was new to me.

jameslamb (approved):
This looks great to me, thanks for the test cases.

I pulled this and installed it, and tried the test case from `cudf` that led to this:

```shell
    --output conda \
    --file-key test_python \
    --matrix "cuda=12.2;arch=aarch64;py=3.10;dependencies=oldest"
```

Saw that produce output with `cupy==12.2.0`, meaning it matched the first matching entry in https://github.com/rapidsai/cudf/blob/9b4c4c721c399bae9e88733da79daa1a10644481/dependencies.yaml#L715-L724, exactly as we'd expect.

bdice:
Thanks for the review @jameslamb! I'm going to merge this and deploy it so that we can get working nightlies for cudf branch-24.10 before code freeze.

PR #113, opened 2024-10-09
fix: add host dependency on 'setuptools', ignore renovate Python updates

jameslamb:
The `renovate` bot just opened #112, trying to upgrade the CI image used in this repo for wheel builds / tests from `python:3.9` to `python:3.13`.

 We don't want those types of updates... it's an intentional choice that `rapids-dependency-file-generator` wheels are built and tested on the oldest Python version this tool supports.

This PR attempts to tell `renovate` not to make such recommendations in the future. I saw "attempts" because I have no idea how to test this.

references suggesting this `ignoreDeps` field is the right way to do this:

* https://docs.renovatebot.com/configuration-options/#ignoredeps
* https://github.com/rapidsai/docker/pull/643#issuecomment-2168645396

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -15,6 +15,7 @@ requirements:
   host:
     - python >=3.9
     - python-build
+    - setuptools
   run:
     - python >=3.9
     - pyyaml
--- a/renovate.json
+++ b/renovate.json
@@ -1,4 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["config:base"]
+  "extends": ["config:base"],
+  "ignoreDeps": [
+    "python"
+  ]
 }


jameslamb:
`conda` builds here failed like this:

```text
ModuleNotFoundError: No module named 'setuptools'
```

([build link](https://github.com/rapidsai/dependency-file-generator/actions/runs/11259576187/job/31308868432?pr=113))

Makes sense... this project uses `setuptools` as its build backend but doesn't declare a host dependency on it in its conda recipe. We must have just been getting that transitively before.

https://github.com/rapidsai/dependency-file-generator/blob/d38a5d3ab21ea13dc4678635dfe5e038fc36ccb1/pyproject.toml#L1-L3

I just pushed a commit here adding a host dependency on `setuptools`, to fix that. I also changed the title to `fix:` so this will result in publishing a new patch version (`1.15.1`)... I think that's warranted, given that this now potentially changes the build environment and therefore the conda package contents.

KyleFromNVIDIA (approved):

PR #115, opened 2024-10-17
feat: support merging output from multiple file keys when writing to stdout

### LINKED ISSUE #114: Support generating multiple file keys at once
We should pass `nargs` parameter to make the `--file-key` CLI arg support generating a file with multiple keys. That will make it much easier for us to create complete environments for cases like unisolated wheel builds where we need multiple sets of dependencies installed into the same environment.

jameslamb:
Closes #114

Adds support for passing `--file-key` multiple times to the same invocation of `rapids-dependency-file-generator`, for `requirements` and `conda` output. This allows writing to stdout the union of multiple file keys.

Like this

```shell
  --output requirements \
  --file-key "py_build_libcuspatial" \
  --file-key "py_rapids_build_libcuspatial" \
  --matrix "cuda=11.8;arch=x86_64;py=3.11;cuda_suffixed=true"
| tee ./requirements-build.txt
```

## Notes for Reviewers

The motivating use case for this is generating `requirements.txt` files containing all the build dependencies for wheels (https://github.com/rapidsai/cuspatial/pull/1473#pullrequestreview-2373307028).

### Design choices?

See inline comments on the diff.

### This does not support `pyproject` output

I cannot think of an application where we'd want to write the merged output of 2 `pyproject.toml` files to stdout, and the implementation to support that would be a lot more difficult than supporting `conda` and `requirements`, because with `pyproject` support we're splicing into an existing file's content (instead of creating the entire file content from scratch).

Here I'm proposing just raising an exception if someone attempts to pass multiple `--file-key` + `--output pyproject` and write to stdout.

### How I tested this

Added new unit tests here.

Also tried some different invocations locally, in `cudf` and `cuspatial`.

Temporarily added code in `ci/build_libcuspatial_wheel.sh` over in https://github.com/rapidsai/cuspatial/pull/1473 to install `rapids-dependency-file-generator` from this branch, and saw it do the right thing there, generating a single list that merged the build backend (`rapids-build-backend` + `scikit-build-core`) and build-time dependencies (e.g. `librmm`):

```text
# This file is generated by `rapids-dependency-file-generator`.
# To make changes, edit . and run `rapids-dependency-file-generator`.
--extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
--extra-index-url=https://pypi.nvidia.com/
cmake>=3.26.4,!=3.30.0
libcudf-cu11==24.12.*,>=0.0.0a0
librmm-cu11==24.12.*,>=0.0.0a0
ninja
rapids-build-backend>=0.3.0,<0.4.0.dev0
scikit-build-core[pyproject]>=0.10.0
wheel
```

([build link](https://github.com/rapidsai/cuspatial/actions/runs/11388486537/job/31685653041?pr=1473#step:9:52))


src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py
```diff
@@ -438,18 +483,50 @@ def make_dependency_files(
...
+        #
+        err_msg = (
+            "Exactly 1 output type should be provided when asking rapids-dependency-file-generator to write to stdout. "
+            "If you see this, you've found a bug. Please report it at https://github.com/rapidsai/dependency-file-generator/issues."
+        )
+        assert output is not None, err_msg
+
+        contents = make_dependency_file(
+            file_type=output.pop(),
+            conda_env_name="rapids-dfg-combined",
```
jameslamb, line 524: The `name: ` field in these generated conda environments is ignored in every workflow I'm aware of in RAPIDS... it's always overwritten by `-n` / `--name` passed through like this:

```shell
  --output conda \
  --file-key test_cpp \
  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch)" | tee "${ENV_YAML_DIR}/env.yaml"

mamba env create --yes -f "${ENV_YAML_DIR}/env.yaml" -n test
```

That's why I'm proposing a hard-coded mildly-informative name instead of taking on the complexity of other alternatives, like:

* allowing `conda_env_name=None` to be passed to `make_dependency_file()`
* trying to come up with some other representative environment name based on which `--file-key`s were passed

src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py
```diff
@@ -438,18 +483,50 @@ def make_dependency_files(
...
+            file_name="ignored-because-multiple-pyproject-files-are-not-supported",
```
jameslamb, line 525: `make_dependency_file()` is already really 3 distinct codepaths inside one function. The codepath for generating `pyproject` output rightly expects to be given a path to an existing `pyproject.toml` file to read in.

I'm proposing just not allowing multiple `--file-key` together with `--output pyproject` and passing through this hard-coded nonsense string instead of taking on the complexity of other options like:

* breaking `make_dependency_file()` up into e.g. `make_conda_env_file()`, `make_requirements_file()`, and `make_pyproject_file()`
* allowing `file_name=None` to be passed to `make_dependency_file()` and generating TOML content from scratch if it is

tests/test_cli.py
```diff
@@ -130,3 +130,47 @@ def test_validate_args():
             "all",
         ]
     )
+
```
jameslamb, line 133: If you are wondering "why isn't there a test case here checking that passing multiple `--file-key` together with `--output pyproject` is rejected?"... it's because I put the exception-raising code for that down further, in `make_dependency_file()`, so that the exception would also be raised if `rapids-build-backend` (which does not use the CLI) tried to pass inputs like that.

I did add a unit test in `test_rapids_dependency_file_generator` checking that that exception is raised as expected.

src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py
jameslamb: Ok, I just pushed https://github.com/rapidsai/dependency-file-generator/pull/115/commits/df3848dd98d51c8aabbdad6adfcb6bfba75c1ea7 changing this to allow `None`.

> *I don't think conda environment files require a name?*

Correct.

```shell
cat > env.yaml <<EOF
channels:
- conda-forge
dependencies:
  - pandas
  - pip
  - pip:
    - scikit-learn
EOF

conda env create \
  --name delete-me \
  --file ./env.yaml
```

Works without issue

```
...
done
#
# To activate this environment, use
#
#     $ conda activate delete-me
#
# To deactivate an active environment, use
#
#     $ conda deactivat
```

tests/test_rapids_dependency_file_generator.py
```diff
@@ -132,6 +147,64 @@ def test_make_dependency_files_should_choose_correct_pyproject_toml(capsys):
...
+    #  * should contain the union of all dependencies from all requested file keys
+    #
+    assert env_dict["dependencies"] == [
+        "matplotlib",
+        "pandas<3.0",
+        "pip",
+        "scikit-learn>=1.5",
+        {"pip": [
+            "folium",
+            "numpy >=2.0",
```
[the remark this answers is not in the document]
jameslamb: I removed the spaces in https://github.com/rapidsai/dependency-file-generator/pull/115/commits/df3848dd98d51c8aabbdad6adfcb6bfba75c1ea7

> *More broadly, do we think DFG should be in the business of normalizing any of this kind of thing on output?*

I'm indifferent about this. I do think it'd probably be *safe* to have DFG do `dep.replace(" ", "")` unconditionally. And there are benefits... slightly smaller output, slightly less work for solvers to do.

But I definitely would prefer it not be part of the scope of this PR.

src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py
jameslamb: Let me try a different way with links.

After processing `dependencies.yaml`, filtering stuff by output type and matrices, deduplicating, etc., DFG wants the literal text that it'll either echo to `stdout` or write to a file.

For that, it calls `make_dependency_file()`:

https://github.com/rapidsai/dependency-file-generator/blob/6e1081109351711c2c81c271861e0b94eb305edd/src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py#L439

Despite the name, that function actually just creates and returns that text... it never writes anything to the filesystem.

The body of the function, in pseudocode, is like:

```python
def make_dependency_file(...):
    relative_path_to_config_file = os.path.relpath(config_file, output_dir)
    file_contents = textwrap.dedent(
        f"""\
        {HEADER}
        # To make changes, edit {relative_path_to_config_file} and run `{cli_name}`.
        """
    )
    if conda:
        # {code specific to conda env YAML files}
    elif requirements:
        # {code specific to requirements.txt}
    elif pyproject:
        # {code specific to pyproject.toml}

    return file_contents
```

So there's very little shared code in there. The list of arguments for the function contains a mix of things that are only used for some but not all of the output types.

Like this:

https://github.com/rapidsai/dependency-file-generator/blob/6e1081109351711c2c81c271861e0b94eb305edd/src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py#L117-L119

That's complex but has been kind of "fine" as long as this was always being called with the data from a single `--file-key`.

For this PR, we now want to use that function to generate a type out of output from multiple `--file-key` entries. For that, there's no single representative `output_dir`, for example. Here, I've just kind of awkwardly provided values with the right types so `mypy` will be happy, with the knowledge that they don't really matter (since most of the arguments are `pyproject`-specific, and this PR isn't supporting pyproject.toml).

It'd be clearer if `make_dependency_file()` was decomposed into `make_conda_env_file()`, `make_requirements_file()` and `make_pyproject_file()`. But I really really did not want to take that on in the scope of this PR.

tests/test_rapids_dependency_file_generator.py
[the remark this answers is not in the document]
bdice: For sure not in scope here. Thanks for considering. We can think about it some other time if it becomes a significant pain point.

src/rapids_dependency_file_generator/_rapids_dependency_file_generator.py
```diff
@@ -95,27 +96,33 @@ def grid(gridspec: dict[str, list[str]]) -> Generator[dict[str, str], None, None
...
 
     Parameters
     ----------
     file_type : Output
         An Output value used to determine the file type.
-    name : PathLike
-        The name of the file to write.
+    conda_env_name : str | None
+        Name to put in the 'name: ' field when generating conda environment YAML files.
+        If ``None``, the generated cond environment file will not have a 'name:' entry.
```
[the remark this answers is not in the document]
jameslamb: blegh thank you!

PR #119, opened 2024-12-08
feat: Allow null matrix entries.

bdice:
This PR allows null matrix entries to co-exist with non-null matrix entries.

Currently a matrix like
```
files:
  all:
    output: conda
    matrix:
      cuda: ["11.8", "12.5"]
      arch: [x86_64]
```

implies that undefined variables like `py:` are ignored.

With this PR, you can add a line like

```
files:
  all:
    output: conda
    matrix:
      cuda: ["11.8", "12.5"]
      arch: [x86_64]
      py: [null, "3.13t"]
```

to generate the same environments as before (where `py` was undefined) and additionally generate environments where `py: "3.13t"` is set.

I retained the naming behavior so that if a key is `null`, it does not appear in the environment name (we get `all_cuda-125_arch-x86_64.yaml` instead of `all_cuda-125_arch-x86_64_py-None.yaml` because I thought the `py-None` was an unwanted change).

An alternative to this would be to specify a different file key where the new matrix key is defined, but that would require replicating the `includes:` list.


vyasr:
@bdice is this ready for a review?

bdice:
@vyasr Yes, it is ready for review.

vyasr (approved):
Seems fine to me. Agree with the choice of retaining the current naming scheme.

PR #129, opened 2025-02-27
Replace `conda-mambabuild` with `conda-build`

jakirkham:
These are now equivalent in behavior. However support for `conda-mambabuild` is being dropped. So switch to `conda-build`.

xref: https://github.com/rapidsai/build-planning/issues/149

### DIFF
--- a/ci/build-test/conda.sh
+++ b/ci/build-test/conda.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 ./ci/update-versions.sh "${RELEASE_VERSION:-}"
 
-mamba install -y boa conda-build
-conda mambabuild \
+mamba install -y conda-build
+conda build \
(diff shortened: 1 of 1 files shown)


ci/build-test/conda.sh
```diff
@@ -3,7 +3,7 @@ set -euo pipefail
 
 ./ci/update-versions.sh "${RELEASE_VERSION:-}"
 
-mamba install -y boa conda-build
-conda mambabuild \
+mamba install -y conda-build
+conda build \
```
jakirkham, line 7: As this was installing `boa` itself, changes in the CI image would not have affected it. Still given `boa` is deprecated, it would be good to clean this up

xref: https://github.com/mamba-org/boa/pull/403

jameslamb (approved):

jakirkham:
Thanks James! 🙏

PR #134, opened 2025-05-22
chore(ci): update pre-commit.ci commit messages

jameslamb:
This project enforces some conventions on commit messages (including PR titles, which become commit messages on squash-merge), which are used to automate versioning.

The automatic `pre-commit.ci` PRs like #133 don't currently generate titles that comply with that (I manually changed the title of that one to get it working).

This proposes modifying the configuration for that service, so its PRs will be mergeable without manual renaming effort. For reference, see https://pre-commit.ci/#configuration


gforsyth (approved):
:shipit: 

conventional commits all the way

PR #138, opened 2025-06-10
chore(deps)!: Remove dask from default label list

vyasr:
Contributes to https://github.com/rapidsai/build-planning/issues/166

### DIFF
--- a/src/rapids_dependency_file_generator/_constants.py
+++ b/src/rapids_dependency_file_generator/_constants.py
@@ -4,7 +4,6 @@ default_channels = [
     "rapidsai",
     "rapidsai-nightly",
     "conda-forge",
-    "nvidia",
 ]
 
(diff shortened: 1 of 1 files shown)


src/rapids_dependency_file_generator/_constants.py
```diff
@@ -3,7 +3,6 @@
 default_channels = [
     "rapidsai",
     "rapidsai-nightly",
-    "dask/label/dev",
     "conda-forge",
     "nvidia",
```
bdice, line 7: We can also drop `nvidia` which is only needed for CUDA 11. See references here:
- https://github.com/rapidsai/gha-tools/pull/146 for `nvidia`
- https://github.com/rapidsai/gha-tools/pull/157 for `dask/label/dev`
```suggestion
```

bdice (approved):

PR #140, opened 2025-06-18
chore(ci): use 'matplotlib-base' instead of 'matplotlib' in conda examples

jakirkham:
When installing Matplotlib with Conda, use `matplotlib-base`. This is more lightweight as it doesn't pull in a bunch of GUI bits that we don't use.

ref: https://conda-forge.org/docs/maintainer/knowledge_base/#matplotlib

### DIFF
--- a/tests/examples/overlapping-deps/dependencies.yaml
+++ b/tests/examples/overlapping-deps/dependencies.yaml
@@ -57,12 +57,13 @@ dependencies:
           - scikit-learn>=1.5
   test_python:
     common:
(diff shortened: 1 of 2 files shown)


jakirkham (commented):
Thanks James! 🙏

Had looked at that, but was unclear on how Conda and pip are both being handled here

Pushed a change below. Though this is a guess. Happy to revise with your feedback

Also yes was doing this mainly to communicate best practices (should others look for examples in this repo)

jameslamb:
Thanks @jakirkham

PR #185, opened 2026-06-22
feat: add renovate cooldown period

gforsyth:
Adds renovate config to wait for 1 week before bumping dependencies to help mitigate supply chain attacks.

xref rapidsai/build-planning#273


jameslamb (approved):
great yes we should do this everywhere we can

PR #197, opened 2026-08-03
feat: Add `constraints` output type

### LINKED ISSUE #196: New output type: constraints
As seen from https://github.com/rapidsai/cudf/pull/23516, the dependencies in a constraints file are not necessarily the same as the ones in a requirements file. In particular, constraints cannot have extras. We should create a new output type, `constraints`, that is identical to `requirements` in functionality but allows constraints to be separated from requirements and not have the extras that requirements have.

KyleFromNVIDIA:
Because constraints don't allow extras, sometimes we want to separate the generation of requirements and constraints. Add a `constraints` output type that is almost identical to `requirements`.

Fixes https://github.com/rapidsai/dependency-file-generator/issues/196


bdice (approved):

bdice:
This seems alright to me. It would make `dependencies.yaml` more expressive.

src/rapids_dependency_file_generator/_config.py
```diff
@@ -33,6 +33,9 @@ class Output(Enum):
     REQUIREMENTS = "requirements"
     """Generate a ``requirements.txt``."""
 
+    CONSTRAINTS = "constraints"
+    """Generate a ``constraints.txt``."""
```
jameslamb, line 37: Can you please expand the docs in the `README`, to make this option more visible?

Adding `constraints` to `output_types` docs and a working example to the inline example YAML there would help.

tests/test_rapids_dependency_file_generator.py
```diff
@@ -70,17 +70,18 @@ def test_make_dependency_file(mock_relpath):
...
+            file_type=file_type,
+            conda_env_name="tmp_env",
+            file_name="tmp_env.txt",
+            config_file="config_file",
+            output_dir="output_path",
+            conda_channels=["rapidsai", "nvidia"],
+            dependencies=["dep1", "dep2"],
+            extras=None,
+        )
+        assert env == header + "dep1\ndep2\n"
```
jameslamb:
```suggestion
        assert env == header + "dep1\ndep2\n", f"output did not match expectations for file type '{file_type}'"
```

In this `for` loop format, I don't think the assertion error or traceback would tell us whether the failure was for `requirements` or `constraints`. Could you add a message to help with that? Think it would reduce the time to debug failures.

